Issue #4, opened 2025-03-10
Introduce a test suite to ensure scale factors up to 1 generate truly compatible data.


clflushopt:
The goal here is to have a full test suite that ensures small scale factor increments up to 1 match 1-1 what we expect, the process is rather easy here. Use the Trino patch in `/patches` to generate some datasets for several small factor increments up to 1 then write a Rust suite that reads the files and compares the outputs to what we already have.

clflushopt:
We already have a test suite for small scale factors so now we need to scale it up to 1. @alamb I am wondering if it makes more sense to have the test suite match the `dbgen` output i.e. (not use CSV instead use `|` as a separator) ?

alamb:
> We already have a test suite for small scale factors so now we need to scale it up to 1. [@alamb](https://github.com/alamb) I am wondering if it makes more sense to have the test suite match the `dbgen` output i.e. (not use CSV instead use `|` as a separator) ?

That is what I would suggest personally

Issue #3, opened 2025-03-10
Introduce unit tests for small scale factors using sampled data.


clflushopt:
We can use a small sample from Apache Trino's implementation to get some unit tests for all the generators.

clflushopt:
I just pushed my patch to Apache Trino's original implementation in my fork https://github.com/clflushopt/tpch this way we can keep a copy of the patched version that will be used to seed our test suite.

clflushopt:
Instead of sampling data and unit tests I am opting for just integration tests with end to end testing instead. Seems more appropriate and less likely to be subject to biased sample picking.

PR #11, opened 2025-03-12
feat: Introduce gzipped test fixtures and steps to reproduce.

clflushopt:
In order to build a nice test suite we need some initial fixtures that can be easily reproduced. In this change I add two sets of data files that I generated from the Java implementation for small scale factors (0.01 + 0.001) the idea is to have a small set of test data that we can easily build our test suite against.

In the change are two patch files that explain how to reproduce the data itself (this is important since we should aim to have a reproducible flow).

This PR is the first step to address https://github.com/clflushopt/tpchgen-rs/issues/4 (I will have a separate PR for https://github.com/clflushopt/tpchgen-rs/issues/3), since we still want a tiered package of tests (unit for fast iteration and integration for ensuring compatibility).

@alamb FYI


alamb:
Thanks @clflushopt  

>  In this change I add two sets of data files that I generated from the Java implementation for small scale factors (0.01 + 0.001) the idea is to have a small set of test data that we can easily build our test suite against.

I will try and do the same thing with the original c implementation just to check that they match

alamb:
Hi @clflushopt -- I compared the results that you checked in in this PR with the results from dbgen (the orignal C one) for both scale factors and I found one discrepancy
- https://github.com/clflushopt/tpchgen-rs/issues/14


I also made a PR with instructions that explain what I did:
- https://github.com/clflushopt/tpchgen-rs/pull/13

Issue #9, opened 2025-03-10
Implement a `dbgen` compatible command line API


alamb:
The tpch data generator also includes a command line utility:

For example, to make scale factor 0.1 (100M):
```shell
 ./dbgen -f -s 0.1
```

The code and flag meaning is on https://github.com/databricks/tpch-dbgen

Here are some examples if its use from the datafusion repo: 

```shell
          mkdir -p datafusion/sqllogictest/test_files/tpch/data
          git clone https://github.com/databricks/tpch-dbgen.git
          cd tpch-dbgen
          make
          ./dbgen -f -s 0.1
          mv *.tbl ../datafusion/sqllogictest/test_files/tpch/data
```

https://github.com/apache/datafusion/blob/9382add72b929c553ca4976d1423d8ebbc80889d/benchmarks/bench.sh#L358-L357

clflushopt:
That's next after I add a test suite thanks for opening the issue I forgot to add it for tracking !

alamb:
I would love to help with this -- and I will with wahtever is needed once I have made  a bit more progress on 

- https://github.com/apache/datafusion/issues/14914

my current obsession: 
- https://github.com/apache/datafusion/issues/15021

clflushopt:
Closed in #12 any further work to improve the UX will be via a separate issue.

PR #12, opened 2025-03-12
feat: scaffold basic dbgen like api

clflushopt:
Scaffold a CLI experience that matches `dbgen`, while we don't have to match it 100% we will strive for conformance so we can ideally be used as a drop in replacement.

- related to https://github.com/clflushopt/tpchgen-rs/issues/9

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -14,32 +14,270 @@
 // The main function is the entry point for the CLI and it uses the `clap` crate
 // to parse the command line arguments and then generate the data.
 
-use clap::{command, Parser};
+// tpchgen-cli/src/main.rs
+use clap::{Parser, ValueEnum};
+use std::fs::{self, File};
+use std::io::{self, BufWriter, Write};
+use std::path::PathBuf;
+use tpchgen::generators::{
+    CustomerGenerator, LineItemGenerator, NationGenerator, OrderGenerator, PartGenerator,
+    PartSupplierGenerator, RegionGenerator, SupplierGenerator,
+};
 
-#[derive(Debug, Parser)]
-#[command(version, about = "TPCH data generation CLI with a dbgen compatible API.", long_about = None)]
-struct Args {
-    #[arg(
-        short,
-        long,
-        help = "Scale factor to use can be one of (0.01, 0.1, 1, 10, 30, 100, 300, 1000, 3000"
-    )]
-    scale: f64,
+#[derive(Parser)]
+#[command(name = "tpchgen")]
+#[command(about = "TPC-H Data Generator", long_about = None)]
+struct Cli {
+    /// Scale factor to address defaults to 1.
+    #[arg(short, long, default_value_t = 1)]
+    scale_factor: i64,
 
-    #[arg(
-        short,
-        help = "Generate data for only the specified tables (defaults to all)."
-    )]
-    tables: Vec<String>,
+    /// Output directory for generated files
+    #[arg(short, long)]
+    output_dir: PathBuf,
 
-    #[arg(short, long, help = "Output format for the data (CSV or Parquet)")]
-    format: String,
+    /// Which tables to generate (default: all)
+    #[arg(short, long)]
+    tables: Option<Vec<Table>>,
 
-    #[arg(short, long, help = "Output directory for the generated data")]
-    output: String,
+    /// Number of parts to generate (for parallel generation)
+    #[arg(short, long, default_value_t = 1)]
+    parts: i32,
+
+    /// Which part to generate (1-based, only relevant if parts > 1)
+    #[arg(long, default_value_t = 1)]
+    part: i32,
 }
 
-fn main() {
-    let args = Args::parse();
-    println!("Hello, world!");
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+enum Table {
+    Nation,
+    Region,
+    Part,
+    Supplier,
+    PartSupp,
+    Customer,
+    Orders,
+    LineItem,
+}
+
+fn main() -> io::Result<()> {
+    // Parse command line arguments
+    let cli = Cli::parse();
+
+    // Create output directory if it doesn't exist
+    fs::create_dir_all(&cli.output_dir)?;
+
+    // Determine which tables to generate
+    let tables: Vec<Table> = if let Some(tables) = cli.tables.as_ref() {
+        tables.clone()
+    } else {
+        vec![
+            Table::Nation,
+            Table::Region,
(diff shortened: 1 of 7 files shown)


clflushopt:
This is missing a rebase so there are some unrelated extra changes.

clflushopt:
@alamb let me know if you have any feedback on this up to now.

tpchgen-cli/src/main.rs
```diff
@@ -1,3 +1,283 @@
...
+#[command(name = "tpchgen")]
+#[command(about = "TPC-H Data Generator", long_about = None)]
+struct Cli {
+    /// Scale factor to address defaults to 1.
+    #[arg(short, long, default_value_t = 1)]
+    scale_factor: i64,
+
+    /// Output directory for generated files
+    #[arg(short, long)]
+    output_dir: PathBuf,
```
alamb, line 37: Maybe we can default to `.` (which is what dbgen does 🤮 )

tpchgen-cli/src/main.rs
```diff
@@ -1,3 +1,283 @@
...
+    PartSupplierGenerator, RegionGenerator, SupplierGenerator,
+};
+
+#[derive(Parser)]
+#[command(name = "tpchgen")]
+#[command(about = "TPC-H Data Generator", long_about = None)]
+struct Cli {
+    /// Scale factor to address defaults to 1.
+    #[arg(short, long, default_value_t = 1)]
+    scale_factor: i64,
```
alamb, line 33: It would be great to accept a `f64` here to test scale factor 0.01, etc

alamb (approved):
This is great -- thank you @clflushopt 

I think changing the options types would help make this a bit easier to use but otherwise 💯 


```shell
mkdir /tmp/tpchdbgen-rs
cargo run --release -- -s 0.001 --output-dir=/tmp/tpchdbgen-rs

```

And compared the output of running 
```
mkdir /tmp/dbgen
docker run -v /tmp/dbgen:/data -it  ghcr.io/scalytics/tpch-docker:main -vf -s 0.001
```

It was the same byte for byte 🏅 
```
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ diff /tmp/tpchdbgen-rs/nation.tbl /tmp/tpchdbgen-rs/nation.tbl
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ diff /tmp/tpchdbgen-rs/region.tbl /tmp/tpchdbgen-rs/region.tbl
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ diff /tmp/tpchdbgen-rs/orders.tbl  /tmp/tpchdbgen-rs/orders.tbl
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ diff /tmp/tpchdbgen-rs/customer.tbl  /tmp/tpchdbgen-rs/customer.tbl
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ diff /tmp/tpchdbgen-rs/lineitem.tbl  /tmp/tpchdbgen-rs/lineitem.tbl
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ diff /tmp/tpchdbgen-rs/part.tbl  /tmp/tpchdbgen-rs/part.tbl
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ diff /tmp/tpchdbgen-rs/partsupp.tbl  /tmp/tpchdbgen-rs/partsupp.tbl
andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$
```


BTW  tpchdbgen-rs takes

```
real	0m17.362s
```

While dbgen takes
```
real	0m46.738s
```

Already pretty 🌶️  !

tpchgen-cli/src/main.rs
clflushopt: Will fix.

tpchgen-cli/src/main.rs
clflushopt: Will fix, set the default to the current directory.

clflushopt:
Thanks @alamb for taking the time to check and review this, performance wise I haven't started making those changes but now that we have the test suite up and running I am going to be more comfortable making some changes there. My goal is to have something at least as fast as the existing Java implementation.

alamb:
> Thanks @alamb for taking the time to check and review this, performance wise I haven't started making those changes but now that we have the test suite up and running I am going to be more comfortable making some changes there. 

Awesome!

> My goal is to have something at least as fast as the existing Java implementation.

I hope this is not a high bar (we can make this really fast)

> One thing I'd like your input on is what format to use by default, I feel like keeping the existing format `dbgen` compatible in the tests would be better instead of opting for using a CSV like one.

I would definitely suggest keeping the dbgen compatible `.tbl` format (BTW it can be thought of as "CSV" with a different separator character (`|`)


> This way we can make it easier to run the checks against both the default fixtures of Trino's implementation and `dbgen` without much changes. Of course I am going to add support accepting a `Sink<T>` in the API so library users can have the option of writing to other formats (Parquet comes to mind here).

Yes, 100% this would be great. As a first step I suggest we have an API that we can use to generate Arrow record batches directly (those can be fed to parquet as well as be processed directly by many other systems)

To make the conversion *really* fast I think we will need to be able to generate one column at a time (where the current code  does it a row at a time)


So instead of something like this

```rust
    for part in generator.iter() {
        writeln!(
            writer,
            "{}|{}|{}|{}|{}|{}|{}|{:.2}|{}",
            part.p_partkey,
            part.p_name,
            part.p_mfgr,
            part.p_brand,
            part.p_type,
            part.p_size,
            part.p_container,
            part.p_retailprice,
            part.p_comment
        )?;
    }
```

Something more like this 

```rust
while !generator.empty() {
   // next 8000 rows of part_key
   let part_key: StringArray = generator.iter().take(8000).map(|p| p_partkey)).collect();
   // next 8000 rows of part_key
   let p_name: StringArray = generator.iter().take(8000).map(|p| p_name)).collect();
   // ...  for other olumns
   let batch = RecordBatch::try_new(vec![part_key, p_name, ...]))
}
....
```

clflushopt:
Thanks for the feedback @alamb I will make the adjustments necessary to ensure we have a `dbgen` compatible output and I'll make a separate issue to switch to a columnar model as well (I think I might get it all done over the weekend).

clflushopt:
@alamb so to make it easier for both of us to test against `dbgen` I've added a bash script that runs what I called "conformance tests" between our rust implementation and `dbgen` hopefully this streamlines your workflow too !

clflushopt:
I am going to merge this and then open a second pull request to improve exclusively on UX.

PR #13, opened 2025-03-12
Add instructions for verifying checked in results

alamb:
- Related to  https://github.com/clflushopt/tpchgen-rs/pull/11

Here are some instructions on how to reproduce and compare the results of the data generator. I have found a few discrepancies with this method that I will file as a follow on ticket


clflushopt:
Thanks @alamb super appreciated ! Now we have another data point to check conformance.

PR #16, opened 2025-03-13
feat: Introduce end-to-end conformance integration tests.

### LINKED ISSUE #3: Introduce unit tests for small scale factors using sampled data.

clflushopt:
Closes #3 and #4

This change introduces end to end conformance tests against the pre-existing features in the data folder that we generated from the Java implementation.

With these tests we can be more comfortable doing performance optimizations and changing the overall generator architecture.

The tests run against fixtures with scale factors of 0.001 and 0.01 and are more than enough to ensure determinism (IMO) I will eventually add 0.1 and 1 scale factor fixtures so we can make sure that we are data compatible with the Java and C original implementations.

To make the testing work with `CARGO_MANIFEST_DIR` to read fixtures I moved the data folder into the `tpchgen` subpackage.

One fun thing with these tests is that I found out that a premature optimization I did originally was indeed buggy (the price in cents is derived from an `i64` computation and found out that some numbers were off by 4 cents lol).


clflushopt:
@alamb let me know if you have any suggestions or improvements that I can make here, one thing I noticed is that maybe I should just ship `dbgen` compatible fixtures i.e. `.tbl` files with `|` as separator instead of doing the `toCSV` thing in my Java patches.

tpchgen/tests/integration_tests.rs
```diff
@@ -0,0 +1,391 @@
...
+        assert_eq!(
+            reference, generated,
+            "Record {} doesn't match for {}.\nReference: {}\nGenerated: {}",
+            i, reference_path, reference, generated
+        );
+    }
+}
+
+#[test]
+fn test_nation_sf_0_001() {
```
alamb, line 66: This is really nice, easy to read code

PR #17, opened 2025-03-14
POC: Optimizations to make data generation faster (3-7x faster)

alamb:
- Draft as it builds on https://github.com/clflushopt/tpchgen-rs/pull/12
- part of https://github.com/clflushopt/tpchgen-rs/issues/6

## Why

I couldn't help myself  😎  (While testing the `dbgen` command line code I wanted to make it faster)

## Performance Results

On my M3 Mac:

| Scale Factor | Main | This PR | times faster |
|--------|--------|--------|--------|
| 0.001 | 17.6s | 2.5s | 7 |
| 1 | 24.1s | 7.95 | 3 |

Note I only applied the "don't copy strings" technique to lineitem. The same thing can be done to all the other tables to get better performance

```
time cargo run --release -- -s 0.001 --output-dir=/tmp/tpchdbgen-rs
...
real	0m17.596s
```

# Background

Some quick profiling on main shows large amounts of time initializing the TextPool and string manipulaton

Screenshot 2025-03-14 at 6 50 33 AM

## Changes
1. Initialize TextPool once, rather than for each table, in 15f3002179d8ca1267b9000829e1b0ded870763e
2. Avoid copying strings in LineItem (use `&str` instead), in c529cf5359b0b76a28c077abfc53d0d714b05700

Here is the profiling after these changes:

Screenshot 2025-03-14 at 8 22 38 AM

Note: I think we could reduce more of the time by formatting the date strings directly into the output (rather than into directly into buffers and then output):

Issue #6, opened 2025-03-10
Improve single threaded performance


clflushopt:
Before tackling the multi-threaded performance, which will probably work out of the box, we should aim to saturate the single thread performance first:

- Use `once_cell` and `lazy_static` to initialze the default distributions and text pools.
- Transverse iterator ownership so we can keep their lifetimes self-contained.
- Use zero-copy wherever we can.

alamb:
Some other ideas

One specific zero copy thing we could do is change `String` to `Arc<String>` here 

https://github.com/clflushopt/tpchgen-rs/blob/fea73ebdfd0fe9c57b78d3957c56f251d106ab87/tpchgen/src/text.rs#L318

I think it would be lots of fun.

alamb:
Applying some of the ideas in this ticket resulted in 3x to 7x speedup (and there is still more lo hanging fruit):
- https://github.com/clflushopt/tpchgen-rs/pull/17

Once https://github.com/clflushopt/tpchgen-rs/pull/12 is merged, I can break it into some smaller PRs for review

clflushopt:
I don't think you need to break it into smaller PRs I can probably go over it and merge it once I merge #12

alamb:
> I don't think you need to break it into smaller PRs I can probably go over it and merge it once I merge [#12](https://github.com/clflushopt/tpchgen-rs/pull/12)

Thanks @clflushopt  -- however I still think separate PRs will be better because:

1.  the changes will be easier to review and 
2. they are independent of #12  (they don't need to be stacked)

Here is the first one: 
- https://github.com/clflushopt/tpchgen-rs/pull/19

alamb:
Here is my plan for optimizating (breaking up what is in https://github.com/clflushopt/tpchgen-rs/pull/17)
- [ ] Preparations (change generators so they return `&str` rather than `String`)
- [ ] Postpone formatting dates as Strings until output
- [ ] Port `LineItem` generator to use references rather than `String`
- [ ] Port `PartSupp` generator to use references rather than `String`
- [ ] Port `Customer` generator to use references rather than `String`
- [ ] Port `Supplier` generator to use references rather than `String`
- [ ] Port `Part` generator to use references rather than `String`
- [ ] Port `Nation` generator to use references rather than `String`
- [ ] Port `Region` generator to use references rather than `String`

First PRs coming up shortly

clflushopt:
[...]

PR #19, opened 2025-03-15
feat: Create `TextPool` once per process rather than once per table

alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/6

# Rationale
- As shown in https://github.com/clflushopt/tpchgen-rs/pull/17, recomputing the TextPool requires several seconds, so doing it once for all tables improves performance significantly

This PR changes
1. Use a single `Arc<TextPool>`  that is initialized once per process
2. Avoid copying the `Vec<u8>` on creation

### DIFF
--- a/tpchgen/src/text.rs
+++ b/tpchgen/src/text.rs
@@ -9,6 +9,7 @@ use crate::{
     distribution::{Distribution, Distributions},
     random::RowRandomInt,
 };
+use std::sync::{Arc, OnceLock};
 
 /// Pool of random text that follows TPC-H grammar.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


tpchgen/src/text.rs
```diff
@@ -190,8 +200,9 @@ impl ByteArrayBuilder {
         self.length
     }
 
-    fn get_bytes(&self) -> Vec<u8> {
-        self.bytes.clone()
+    /// Return the inner bytes and length of the builder.
```
alamb, line 203: This change allows reusing the underlying bytes during `TextPool::new`

clflushopt (approved):
LGTM thanks for this !

PR #20, opened 2025-03-15
feat: Use tbl format for fixtures and tests

clflushopt:
Initially I opted in to use CSV but that turned out to be not so great idea since it required some changes to diff against Java and C fixtures.

I made the changes in the Java fork to just write TBL format compatible files and made the same changes in our Rust integration tests.

Now the outputs are compatible with dbgen and thus can be diff'd directly; same goes for the fixtures I checked in.


alamb:
Awesome!

PR #21, opened 2025-03-15
feat: optimize single threaded performance

clflushopt:
Part of #6 this change introduces several, small optimizations that compound at large scale factors (1+).

- Refactor `RandomStringSequence` to allocate the values vector only once instead of per call to `next_value()` and inline the randomness swapping logic to be on indices to avoid mutating the original vector, below you can see the profile.

**Before**
Screenshot_20250315_135619

**After**
Screenshot_20250315_135606

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,3 +3,6 @@
 members = [ "tpchgen" , "tpchgen-cli"]
 
 resolver = "2"
+
+[profile.release]
+debug = true
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -123,8 +123,7 @@ fn generate_nation(cli: &Cli) -> io::Result<()> {
             nation.n_nationkey, nation.n_name, nation.n_regionkey, nation.n_comment
         )?;
     }
-
-    Ok(())
+    writer.flush()
 }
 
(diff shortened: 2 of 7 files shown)


clflushopt:
@alamb I have a few ideas on how to approach some other similar optimizations especially with regards to memory usage but I'll wait until #17 PR's are merged so we have a new baseline with the speedups you made ! thanks again !

alamb (approved):
I did some performance testing locally:
main: 0m3.288s
this branch:  0m3.158s

However my accuracy is not very high (the numbers are quite variable)

Here is how I tested
```
 time cargo run --release -- -s 0.1   --output-dir=/tmp/tpchdbgen-rs
```

alamb:
@clflushopt  since this will conflict with https://github.com/clflushopt/tpchgen-rs/pull/27 perhaps you can merge this PR and continue with additional optimizations as other PRs?

alamb:
(I can merge it too if you prefer, but wanted to check with you first)

alamb:
> @alamb running with the merged changes is already showing some great perf gains ! I merged (instead of rebasing) and there's no conflict now.

Just wait until we remove Strings copies in LineItem and PartSupp generation!

PR #26, opened 2025-03-17
Change Random* generators to return `&str` or structs rather than `String` (10% faster)

alamb:

The overall plan is described [here](https://github.com/clflushopt/tpchgen-rs/issues/6#issuecomment-2728967340). I am trying to make these PRs easy to review by making them in chunks.

## Rationale: 
Several of the data generators currently return `String` which forces a copy. The data generator then copies the data again into the output immediately, so avoiding the intermediate copy makes things faster. 

## Changes:
1. Update the generators as much as possible to avoid creating `String`s . 
2. Avoid UTF-8 validation on known ASCII data 

This PR actually does make things a bit faster (but most of the data is still turned into a String by the iterators), but in future PRs I will change the iterators so they do not use Strings either. 

I will comment more inline

## Performance
Performance (of `target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs`)
Main: real	0m11.045s
This PR: real	0m10.015s

### DIFF
--- a/tpchgen/src/generators.rs
+++ b/tpchgen/src/generators.rs
@@ -131,7 +131,7 @@ impl Iterator for NationGeneratorIterator {
             // n_regionkey
             n_regionkey: self.nations.get_weight(self.index) as i64,
             // n_comment
-            n_comment: self.comment_random.next_value(),
+            n_comment: self.comment_random.next_value().to_string(),
         };
 
         self.comment_random.row_finished();
@@ -241,7 +241,7 @@ impl Iterator for RegionGeneratorIterator {
         let region = Region {
             r_regionkey: self.index as i64,
             r_name: self.regions.get_value(self.index).to_string(),
-            r_comment: self.comment_random.next_value(),
+            r_comment: self.comment_random.next_value().to_string(),
         };
 
         self.comment_random.row_finished();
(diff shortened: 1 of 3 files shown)


tpchgen/src/generators.rs
```diff
@@ -131,7 +131,7 @@ impl Iterator for NationGeneratorIterator {
             // n_regionkey
             n_regionkey: self.nations.get_weight(self.index) as i64,
             // n_comment
-            n_comment: self.comment_random.next_value(),
+            n_comment: self.comment_random.next_value().to_string(),
```
alamb, line 134: As mentioned in the description, there is still a lot of `String` copying going on, but after this PR is it more explicit.

Follow on PRs will remove these calls to `to_string` and actually make the code faster

tpchgen/src/random.rs
```diff
@@ -299,6 +300,7 @@ pub struct RandomAlphaNumeric {
     inner: RowRandomInt,
     min_length: i32,
     max_length: i32,
+    buffer: Vec<u8>,
```
alamb, line 303: Reuses the same allocation across calls to avoid having to copy bytes again into String

tpchgen/src/random.rs
```diff
@@ -377,16 +380,13 @@ impl RandomPhoneNumber {
...
-        let country_code = 10 + (nation_key % Self::NATIONS_MAX as i64);
-        let local1 = self.inner.next_int(100, 999);
-        let local2 = self.inner.next_int(100, 999);
-        let local3 = self.inner.next_int(1000, 9999);
-
-        format!(
-            "{:02}-{:03}-{:03}-{:04}",
-            country_code, local1, local2, local3
-        )
+    pub fn next_value(&mut self, nation_key: i64) -> PhoneNumberInstance {
```
alamb, line 383: Returning a structure that can be formatted with `{}`  means the code will (eventually) be able to format the phone number directly into the output stream, rather than formatting (copying) first into an intermediate buffer

tpchgen/src/text.rs
```diff
@@ -62,22 +62,14 @@ impl TextPool {
         self.size
     }
 
-    /// Returns a chunk of text from the pool.
-    pub fn text(&self, begin: i32, end: i32) -> String {
-        assert!(begin >= 0, "Begin index must be greater than or equal to 0");
```
alamb, line 67: I removed these asserts as they are redundant in Rust -- by default `&data[begin..end]` will check that `begin..end` are within bounds and `panic` if they are not.

clflushopt (approved):
LGTM this is nice @alamb !

PR #27, opened 2025-03-17
Change date handling to use TPCHDate type rather than String (10% faster)

alamb:
- Similarly to https://github.com/clflushopt/tpchgen-rs/pull/26

The overall plan is described [here](https://github.com/clflushopt/tpchgen-rs/issues/6#issuecomment-2728967340). I am trying to make these PRs easy to review by making them in chunks.

## Rationale: 
Several of the data generators represent Dates using a `String` which forces a copy. The data generator then copies the data again into the output immediately, so avoiding the intermediate copy makes things faster. 

## Changes:
1. Use a TPCHDate type rather than a string to represent dates

## Performance
This PR also improves performance on its own (so when combined with https://github.com/clflushopt/tpchgen-rs/pull/26 we are already significantly better)

Performance (of `target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs`)
Main: real	0m11.045s
This PR: real	0m9.819s

### DIFF
--- a/tpchgen/src/dates.rs
+++ b/tpchgen/src/dates.rs
@@ -1,5 +1,6 @@
 use chrono::NaiveDate;
 use lazy_static::lazy_static;
+use std::fmt::{Display, Formatter};
 
 /// The value of 1970-01-01 in the date generator system
 pub const GENERATED_DATE_EPOCH_OFFSET: i32 = 83966;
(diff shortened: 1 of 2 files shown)


tpchgen/src/dates.rs
```diff
@@ -88,11 +103,6 @@ impl DateUtils {
         result + offset
     }
 
-    /// Format money value (convert to decimal)
```
alamb, line 91: this is unused and seems unrelated to dates, so I removed it

tpchgen/src/generators.rs
```diff
@@ -1203,7 +1203,7 @@ pub struct Order {
     /// Order total price
     pub o_totalprice: f64,
     /// Order date
-    pub o_orderdate: String, // Could use a date type instead
```
alamb, line 1206: @clflushopt seems to have been thinking the same thing when this code was first written.

So maybe this PR just completes a TODO 😆

clflushopt: Indeed, this and other dates were all supposed to be changed to a native date type like you did !

tpchgen/src/dates.rs
```diff
@@ -49,16 +50,30 @@ impl GenerateUtils {
...
 }
 
-pub struct DateUtils;
+/// Represents a date (day/year)
+///
+/// Example display: 1992-01-01
+#[derive(Debug, Clone, PartialEq)]
+pub struct TPCHDate {
+    inner: NaiveDate,
+}
```
clflushopt, line 59: I am pretty sure at this point we don't need `NaiveDate` and we can probably replace the usage here with an `i32` and add the formatting code we might need. Will think about it, since I'd rather reduce the number of dependencies we have.

clflushopt (approved):
LGTM

alamb:
> LGTM, I think I broke your branch by doing a direct merge into it (was too excited to merge it asap !) sorry about that you might need to rebase now.

No worries! I think it should be fixed now

tpchgen/src/dates.rs
alamb: - Filed https://github.com/clflushopt/tpchgen-rs/issues/28 to track

alamb:
Woohoo!

PR #31, opened 2025-03-18
Alamb/lineitem no copy

alamb:
This is a non-working attempt to keep the `Iterator` interface but return references

Specifically trying to compile it results in the following error

```
error: lifetime may not live long enough
    --> tpchgen/src/generators.rs:2021:9
     |
2015 | impl<'a> Iterator for LineItemIterator<'a> {
     |      -- lifetime `'a` defined here
...
2018 |     fn next(&mut self) -> Option<Self::Item> {
     |             - let's call the lifetime of this reference `'1`
...
2021 |         self.generator.next_line_item(advance)
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method was supposed to return data with lifetime `'a` but it is returning data with lifetime `'1`
```

I couldn't figure out how to make it work, so I am going to work on a non iterator based approach

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -20,7 +20,7 @@ use std::fs::{self, File};
 use std::io::{self, BufWriter, Write};
 use std::path::PathBuf;
 use tpchgen::generators::{
-    CustomerGenerator, LineItemGenerator, NationGenerator, OrderGenerator, PartGenerator,
+    CustomerGenerator, LineItemGeneratorBuilder, NationGenerator, OrderGenerator, PartGenerator,
     PartSupplierGenerator, RegionGenerator, SupplierGenerator,
 };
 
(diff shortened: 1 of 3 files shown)


clflushopt:
I have a vague feeling this is similar to the state of `std::io::lines` and requires workarounds or GADTs to solve [1] & [2] (I could be wrong though because I also have a feeling we might get away with it if we restructure the ownership tree better here).

[1]: https://users.rust-lang.org/t/generic-associated-types-and-lifetime/121290/9
[2]: https://blog.rust-lang.org/2022/10/28/gats-stabilization.html
[3]: https://lukaskalbertodt.github.io/2018/08/03/solving-the-generalized-streaming-iterator-problem-without-gats.html

alamb:
@scsmithr  found the way in https://github.com/clflushopt/tpchgen-rs/pull/35

PR #33, opened 2025-03-18
Unbox writer (3-5% faster)

alamb:

### Rationale
Using a `Box<dyn Write>` when there is only one actual writer implementation prevents certain inlining from occurring.

I noticed this while looking at some performance traces

### Changes
Use the actual types for the writer rather than `Box<dyn Write>`

I measured a small but measurable 3-5% improvement using `time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs`

| branch | time |
|--------|--------|
| `main` | 0m8.573s |
| this branch | 0m8.297s |

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -102,12 +102,10 @@ fn main() -> io::Result<()> {
     Ok(())
 }
 
-fn new_table_writer(cli: &Cli, filename: &str) -> io::Result<Box<dyn Write>> {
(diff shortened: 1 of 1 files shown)


clflushopt (approved):
LGTM fyi I did experiment with different buffer capacities for the `BufWriter` I noticed there are speed ups when I increased the buffer to 16 or 32 MBs on my Linux machine but I am not confident it makes sense since it would depend on page sizes and OS buffer caches because the improvements were not noticeable on my Macbook so i didn't add those.

Issue #10, opened 2025-03-10
Implement multi-threading data generation


alamb:
Another thing that would also likely be helpful for generation is to make the code multi-threaded -- so one thread creates the output stream (of `Arc<String>` or whatever) and another thread actually writes them out to CSV files.

It would also be amazing to generate all the tables in parallel (rather than doing it one at a time)

alamb:
> Yes I am planning both (multiple tables in parallel and supporting multiple generators to writers).

I would love to help

alamb:
I have been thinking about "how would we use multiple threads to generate line item data"

What I was thinking about so far is:
1. Make sure all generator / iterators are `Clone` (so you can `.clone()` to get a copy)
2. Add a `skip_num(num_records: usize)` method to each one

Then we could generate output chunks in parallel like
```rust
let iter = LineItemGenerator::new(...).iter();

// make first 1000 rows into a buffer
let chunk_1 = generate_lines(iter.clone().take(1000));
// make next 1000 rows ...
let iter = iter.skip_num(1000); // call special skip which just advances the rngs
let chunk_2 = generate_lines(iter.clone().take(1000)
...
let chunk_N = generate_lines(iter.clone().take(1000)

println!("{}", chunk_1);
println!("{}", chunk_2);
...
println!("{}", chunk_n);
```

I think `skip_num` would be pretty fast as it just advances the random number generators (and could skip all actual row generation logic)

So this solution requires advancing each random number generator 2x, but I think it could be put all the cores to use pretty effectively

clflushopt:
I kind of reached the same conclusion; although I am more inclined to hide `skip_num`  logic here. My line of thought is that library users should transparently be able to have parallel and non-parallel iterators (like how `rayon` exposes `par_iter`). But I'll have to tinker about it.

alamb:
> I kind of reached the same conclusion; although I am more inclined to hide `skip_num` logic here. My line of thought is that library users should transparently be able to have parallel and non-parallel iterators (like how `rayon` exposes `par_iter`). But I'll have to tinker about it.

Perhaps we can start just using the normal iterator [`skip`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip) (which will do a bunch of work that is not strictly necessary). If that turns out to be too slow we can look into a `skip_num` optimization -- it could well be a premature optimization

alamb:
[...]

PR #34, opened 2025-03-18
(POC) Line item generation in parallel

alamb:
POC: Multi-threaded LineItem generation

- part of https://github.com/clflushopt/tpchgen-rs/issues/10

This is a POC showing how we might implement multi-threaded generation with the idea in https://github.com/clflushopt/tpchgen-rs/issues/10#issuecomment-2730451061

## `lineitem` at SF=10. From 37s --> 5s :bowtie: 

```shell
time target/release/tpchgen-cli -s 10 --tables line-item  --output-dir=/tmp/tpchdbgen-rs
```

| branch | time |
|--------|--------|
| main | 0m37.383s |
| this pr | 0m5.215s | 

It can generate lineitem at scale factor 100 in 32 seconds:

```
time target/release/tpchgen-cli --tables line-item -s 100 --output-dir=/tmp/tpchdbgen-rs
Total orders: 150000000, num parts: 15001
Using 15 parallel tasks
wrote 15001 buffers, 79579694556 bytes (2.392097480595112 GB bytes/sec)
Generation complete!

real	0m32.470s
user	5m35.949s
sys	0m14.225s
```

## Tokio vs Rayon
This uses tokio rather than Rayon. 

I tried to figure out something similar to `buffered` with Rayon but I couldn't find any way to limit the number of things that were buffered at once (as in if I have 16 cores, I only want to have at most 16 jobs running in parallel so as to avoid trying to buffer the entire output set)

I want to run the jobs  in parallel like this:

```
job1
job2
job3
...
jobN
```
And process the output in that order too, but also limit the number of jobs that are currently outstanding

### DIFF
--- a/tpchgen-cli/Cargo.toml
+++ b/tpchgen-cli/Cargo.toml
@@ -7,3 +7,7 @@ authors = ["clflushopt", "alamb"]
 [dependencies]
 clap = { version = "4.5.32", features = ["derive"] }
 tpchgen = { path = "../tpchgen" }
+tokio = { version = "1.44.1", features = ["full"]}
+futures = "0.3.31"
+num_cpus = "1.0"
+
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -14,11 +14,17 @@
 // The main function is the entry point for the CLI and it uses the `clap` crate
 // to parse the command line arguments and then generate the data.
 
+use std::collections::VecDeque;
 // tpchgen-cli/src/main.rs
 use clap::{Parser, ValueEnum};
+use futures::StreamExt;
 use std::fs::{self, File};
 use std::io::{self, BufWriter, Write};
 use std::path::PathBuf;
+use std::sync::Mutex;
+use std::time::Instant;
+use tokio::task::JoinSet;
+use tpchgen::dates::GenerateUtils;
 use tpchgen::generators::{
     CustomerGenerator, LineItemGenerator, NationGenerator, OrderGenerator, PartGenerator,
     PartSupplierGenerator, RegionGenerator, SupplierGenerator,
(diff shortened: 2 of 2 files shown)


scsmithr:
Just a heads up, I opened up https://github.com/clflushopt/tpchgen-rs/pull/35 which would conflict quite heavily with this.

I'd really want to see if there's a way to keep the `Arc`s out the generators since it'd make it more challenging to tie the str lifetimes to the text pool/distributions.

tpchgen/src/distribution.rs
```diff
@@ -214,142 +214,148 @@ impl DistributionLoader {
 /// Distributions wraps all TPC-H distributions and provides methods to access them.
 #[derive(Debug, Clone)]
 pub struct Distributions {
-    distributions: IndexMap<String, Distribution>,
+    distributions: IndexMap<String, Arc<Distribution>>,
```
alamb, line 217: The key design of this PR is to make multiple instances of `LineItemGeneratorIterator` which each generate data in parallel for different parts of the dataset

Initially, `clone`ing  `Distribution` consumed most of the time, and so in this PR I updated them to all be `Arc`d 

I think @scsmithr took a similar approach (using lifetimes) in this PR:
- https://github.com/clflushopt/tpchgen-rs/pull/35

alamb: Rebased this PR to use that and it works great!

alamb:
> Just a heads up, I opened up #35 which would conflict quite heavily with this.
> 
> I'd really want to see if there's a way to keep the `Arc`s out the generators since it'd make it more challenging to tie the str lifetimes to the text pool/distributions.

THanks @scsmithr  -- I rebased this PR after merging #35 and it still works great (and is screaming fast)

alamb:
> Thanks again @alamb I highly appreciate the help with bringing this together for an initial release wouldn't have been possible without yours and @scsmithr help. Kudos.

it is my pleasure -- this is going to be awesome

I have actually been dreaming of  tool such as this for the last 15 years (ever since I started messing with TPCH) and it is so awesome to see it coming together

alamb:
BTW I plan to clean this up a bit before making a real PR, I just wanted to prove to myself it could be made to work

PR #32, opened 2025-03-18
Avoid copying `String` when generating `LineItem` (20% faster)

alamb:
- Part of https://github.com/clflushopt/tpchgen-rs/issues/6

My overall plan is described [here](https://github.com/clflushopt/tpchgen-rs/issues/6#issuecomment-2728967340), and I am trying to make PRs easy to review by making them in chunks

### Rationale

Creating `String` is expensive:
1. The bytes must be copied
2. The overhead of allocation/deallocation is substantial.

Thus the fewer `String`s created the faster the data generator goes

Once we are happy with this pattern, I plan to apply the same thing to the other 6 tables

### Changes
Rework how `LineItem`'s are generated to avoid creating / copying`String`

### Performance measurements

Testing with
```shell
time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
```

| branch | time |
|--------|--------|
| `main` | 0m8.579s |
| this PR | 0m6.985s |

### Note on not implementing `Iterator` and `IntoIterator`

I tried several times to keep the `Iterator` trait but could not get the lifetimes to work out. I would be grateful if someone else could give it a look and try. Here is what I did:
- https://github.com/clflushopt/tpchgen-rs/pull/31

PR #35, opened 2025-03-18
Add lifetimes to generators and generated types

scsmithr:
Adds lifetimes to all generators and relevant types to remove many cases of using `String`.

Essentially this ties the lifetimes of all strings that aren't modified to either the `TextPool` or `Distribution`. This also remove the `Arc` around `TextPool` to help accommodate this.

There's still a few `String` instances when `format!` (or other string formatting) is needed. I opted to hold off on changing those in this pr.

### DIFF
--- a/tpchgen/src/generators.rs
+++ b/tpchgen/src/generators.rs
@@ -51,7 +51,7 @@ impl<'a> NationGenerator<'a> {
 }
 
 impl<'a> IntoIterator for &'a NationGenerator<'a> {
-    type Item = Nation;
+    type Item = Nation<'a>;
     type IntoIter = NationGeneratorIterator<'a>;
 
     fn into_iter(self) -> Self::IntoIter {
@@ -61,7 +61,7 @@ impl<'a> IntoIterator for &'a NationGenerator<'a> {
 
 /// The NATION table
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Nation {
+pub struct Nation<'a> {
     /// Primary key (0-24)
     pub n_nationkey: i64,
     /// Nation name
@@ -69,10 +69,10 @@ pub struct Nation {
     /// Foreign key to REGION
     pub n_regionkey: i64,
     /// Variable length comment
-    pub n_comment: String,
+    pub n_comment: &'a str,
 }
 
-impl fmt::Display for Nation {
+impl fmt::Display for Nation<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
@@ -82,14 +82,14 @@ impl fmt::Display for Nation {
     }
 }
 
-impl Nation {
+impl<'a> Nation<'a> {
     /// Create a new `nation` record with the specified values.
-    pub fn new(n_nationkey: i64, n_name: &str, n_regionkey: i64, n_comment: &str) -> Self {
+    pub fn new(n_nationkey: i64, n_name: &str, n_regionkey: i64, n_comment: &'a str) -> Self {
         Nation {
             n_nationkey,
             n_name: n_name.to_string(),
             n_regionkey,
-            n_comment: n_comment.to_string(),
+            n_comment,
         }
     }
 }
@@ -118,7 +118,7 @@ impl<'a> NationGeneratorIterator<'a> {
 }
 
 impl<'a> Iterator for NationGeneratorIterator<'a> {
-    type Item = Nation;
+    type Item = Nation<'a>;
 
     fn next(&mut self) -> Option<Self::Item> {
         if self.index >= self.nations.size() {
@@ -133,7 +133,7 @@ impl<'a> Iterator for NationGeneratorIterator<'a> {
             // n_regionkey
             n_regionkey: self.nations.get_weight(self.index) as i64,
             // n_comment
-            n_comment: self.comment_random.next_value().to_string(),
+            n_comment: self.comment_random.next_value(),
         };
 
         self.comment_random.row_finished();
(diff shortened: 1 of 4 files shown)


scsmithr:
cc @clflushopt and @alamb 

(Also I believe this may supersede https://github.com/clflushopt/tpchgen-rs/pull/32)

tpchgen/src/text.rs
```diff
@@ -30,13 +30,9 @@ impl TextPool {
...
-    pub fn get_or_init_default() -> Arc<Self> {
-        Arc::clone(DEFAULT_TEXT_POOL.get_or_init(|| {
-            Arc::new(Self::new(
-                Self::DEFAULT_TEXT_POOL_SIZE,
-                &Distributions::default(),
-            ))
-        }))
+    pub fn get_or_init_default() -> &'static Self {
+        DEFAULT_TEXT_POOL
+            .get_or_init(|| Self::new(Self::DEFAULT_TEXT_POOL_SIZE, &Distributions::default()))
```
scsmithr, line 35: Removed the Arc, allow us to tie lifetimes to text pools much more easily.

tpchgen/tests/integration_tests.rs
```diff
@@ -19,14 +19,9 @@ fn read_tbl_gz<P: AsRef<Path>>(path: P) -> Vec<String> {
...
-    generator_fn: F,
-    reference_path: &str,
-    scale_factor: f64,
-    transform_fn: impl Fn(T) -> String,
-) where
-    G: Iterator<Item = T>,
-    F: FnOnce(f64) -> G,
+fn test_generator<T, I>(iter: I, reference_path: &str, transform_fn: impl Fn(T) -> String)
+where
+    I: Iterator<Item = T>,
```
scsmithr, line 24: I remove the closure being passed in as all it was doing was creating an iterator using an sf (which was also being passed in). This just accepts an iterator now.

tpchgen/tests/integration_tests.rs
```diff
@@ -61,63 +55,55 @@ fn test_generator<T, G, F>(
...
-        "data/sf-0.001/nation.tbl.gz",
-        0.001,
-        |nation| {
-            format!(
-                "{}|{}|{}|{}|",
-                nation.n_nationkey, nation.n_name, nation.n_regionkey, nation.n_comment
-            )
-        },
-    );
+    let _sf = 0.001;
```
scsmithr, line 58: Kept these in just for consistency with other tests that actually use a scale factor.

tpchgen/src/generators.rs
```diff
@@ -47,13 +46,13 @@ impl NationGenerator {
...
     pub fn iter(&self) -> NationGeneratorIterator {
-        NationGeneratorIterator::new(self.distributions.nations(), &self.text_pool)
+        NationGeneratorIterator::new(self.distributions.nations(), self.text_pool)
     }
 }
 
-impl IntoIterator for NationGenerator {
-    type Item = Nation;
-    type IntoIter = NationGeneratorIterator;
+impl<'a> IntoIterator for &'a NationGenerator<'a> {
```
scsmithr, line 53: I'd argue these `IntoIterator` implementations don't really help much since they just call `iter` (which accepts a reference and isn't consuming).

tpchgen/src/text.rs
clflushopt: The `TextPool` is read-only but I wonder if the lack of `Arc` might cause the compiler to complain if we aim to have multiple generators in different threads sharing it.

clflushopt (approved):
LGTM I left a small comment about the absence of `Arc` making it unsafe to share across threads (it's not the case since the global text pool is read only and has a static lifetime but i don't recall how the compiler behaves in this case). Otherwise approved, since I merged #32 before reviewing this there are a few merge conflicts that need to be addressed.

Thanks for the contribution ! I will add @alamb as reviewer so he's aware of the change.

tpchgen/src/generators.rs
clflushopt: I agree I implemented `IntoIter` here because I thought having `T` and `&T` and `&mut T` might be useful but I highly doubt that.

tpchgen/src/generators.rs
```diff
@@ -8,24 +8,23 @@ use crate::random::RandomBoundedLong;
...
 use crate::random::RowRandomInt;
 use crate::text::TextPool;
-use std::sync::Arc;
 
 use crate::dates::{GenerateUtils, TPCHDate};
 use crate::random::{RandomBoundedInt, RandomString, RandomStringSequence, RandomText};
 
 /// Generator for Nation table data
-pub struct NationGenerator {
+pub struct NationGenerator<'a> {
```
alamb, line 16: I think making the generation multi-threaded will be significantly harder using the approach in https://github.com/clflushopt/tpchgen-rs/pull/34 if there are lifetimes on the generators

However it seems to me that all the distributions are effectively `'&static` (don't change) so maybe instead of putting lifetimes on the generators, we could just make all the references to `TextPool` `&'static TextPool` 🤔

tpchgen/src/text.rs
alamb: It makes sense that trying to mix Arc and references is complicated

I actually did the opposite in https://github.com/clflushopt/tpchgen-rs/pull/34 -- that is put `Arc` on the `Distributions` and `Distribution`

alamb (commented):
Thank you so much @scsmithr -- I think this is totally the right direction (don't copy the text pools)

This PR seems to go fro Arc --> `&` which is more efficient in general 🚀  . However I think it will make it hard to pass the generators to different threads. 

What do you think about combining the best of both words and making all the references `&'static` ? That would remove almost all runtime overhead 🤔

tpchgen/src/generators.rs
```diff
@@ -1353,23 +1356,23 @@ impl OrderGenerator {
     }
 }
 
-impl IntoIterator for OrderGenerator {
-    type Item = Order;
-    type IntoIter = OrderGeneratorIterator;
+impl<'a> IntoIterator for &'a OrderGenerator<'a> {
```
alamb, line 1359: this is clever and a neat way to keep the iterator pattern

alamb (approved):
Actually, I played around with this locally and I believe the generators that come back from `LineItemGenerator::new()` do in fact have the static lifetime

I made this change and it compiles:

```diff
diff --git a/tpchgen-cli/src/main.rs b/tpchgen-cli/src/main.rs
index 4d43c4e..ac79f3e 100644
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -114,7 +114,7 @@ fn generate_nation(cli: &Cli) -> io::Result<()> {
     let filename = "nation.tbl";
     let mut writer = new_table_writer(cli, filename)?;

-    let generator = NationGenerator::new();
+    let generator: NationGenerator<'static> = NationGenerator::new();
     for nation in generator.iter() {
         writeln!(
             writer,
```

Therefore I agree this is a better approach than https://github.com/clflushopt/tpchgen-rs/pull/32 and I will revert that to get this one to merge

alamb:
I am working on a PR to remove the last string copies

PR #37, opened 2025-03-19
Remove more String copying (5% faster)

alamb:
- Part of https://github.com/clflushopt/tpchgen-rs/issues/10

This PR removes some more
1. Consolidates format strings in the cli and tests to use the `Display` impl 
2. Avoids some more String copies by postponing formatting

Performance of running
```shell
time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
```
| branch | time |
|--------|--------|
| main | 0m6.637s |
| this PR | 0m6.265s |

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -147,17 +147,7 @@ fn generate_supplier(cli: &Cli) -> io::Result<()> {
 
     let generator = SupplierGenerator::new(cli.scale_factor, cli.part, cli.parts);
     for supplier in generator.iter() {
-        writeln!(
-            writer,
-            "{}|{}|{}|{}|{}|{:.2}|{}|",
-            supplier.s_suppkey,
-            supplier.s_name,
-            supplier.s_address,
-            supplier.s_nationkey,
-            supplier.s_phone,
-            supplier.s_acctbal,
-            supplier.s_comment
-        )?;
+        writeln!(writer, "{supplier}")?;
     }
     writer.flush()
 }
@@ -179,18 +169,7 @@ fn generate_customer(cli: &Cli) -> io::Result<()> {
 
     let generator = CustomerGenerator::new(cli.scale_factor, cli.part, cli.parts);
     for customer in generator.iter() {
-        writeln!(
-            writer,
-            "{}|{}|{}|{}|{}|{:.2}|{}|{}|",
-            customer.c_custkey,
-            customer.c_name,
-            customer.c_address,
-            customer.c_nationkey,
-            customer.c_phone,
-            customer.c_acctbal,
-            customer.c_mktsegment,
-            customer.c_comment
-        )?;
+        writeln!(writer, "{customer}",)?;
     }
     writer.flush()
 }
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -114,11 +114,7 @@ fn generate_nation(cli: &Cli) -> io::Result<()> {
 
     let generator = NationGenerator::new();
     for nation in generator.iter() {
-        writeln!(
-            writer,
-            "{}|{}|{}|{}|",
-            nation.n_nationkey, nation.n_name, nation.n_regionkey, nation.n_comment
-        )?;
+        writeln!(writer, "{nation}",)?;
     }
     writer.flush()
 }
@@ -129,11 +125,7 @@ fn generate_region(cli: &Cli) -> io::Result<()> {
 
     let generator = RegionGenerator::new();
     for region in generator.iter() {
-        writeln!(
-            writer,
-            "{}|{}|{}|",
-            region.r_regionkey, region.r_name, region.r_comment
-        )?;
+        writeln!(writer, "{region}",)?;
     }
     writer.flush()
 }
@@ -144,19 +136,7 @@ fn generate_part(cli: &Cli) -> io::Result<()> {
 
     let generator = PartGenerator::new(cli.scale_factor, cli.part, cli.parts);
     for part in generator.iter() {
-        writeln!(
-            writer,
-            "{}|{}|{}|{}|{}|{}|{}|{:.2}|{}|",
-            part.p_partkey,
-            part.p_name,
-            part.p_mfgr,
-            part.p_brand,
-            part.p_type,
-            part.p_size,
-            part.p_container,
-            part.p_retailprice,
-            part.p_comment
-        )?;
+        writeln!(writer, "{part}",)?;
     }
     writer.flush()
 }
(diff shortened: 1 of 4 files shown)


tpchgen-cli/src/main.rs
```diff
@@ -114,11 +114,7 @@ fn generate_nation(cli: &Cli) -> io::Result<()> {
 
     let generator = NationGenerator::new();
     for nation in generator.iter() {
-        writeln!(
-            writer,
-            "{}|{}|{}|{}|",
-            nation.n_nationkey, nation.n_name, nation.n_regionkey, nation.n_comment
-        )?;
+        writeln!(writer, "{nation}",)?;
```
alamb, line 117: the idea is to use the `Display` impl which is the same

Then we can change the underlying representation of some fields (e.g instead of `"Customer#10234"` the field stores `10234` and the formatter adds the Customer part)

tpchgen/src/generators.rs
```diff
@@ -293,7 +293,7 @@ impl fmt::Display for Part<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "{}|{}|{}|{}|{}|{}|{}|{:.2}|{}|",
+            "{}|{}|Manufacturer#{}|Brand#{}|{}|{}|{}|{:.2}|{}|",
```
alamb, line 319: moved the formatting into Display rather than creating a String to then copy again into output

tpchgen/tests/integration_tests.rs
```diff
@@ -82,18 +76,7 @@ fn test_part_sf_0_001() {
     let sf = 0.001;
     let generator = PartGenerator::new(sf, 1, 1);
     test_generator(generator.iter(), "data/sf-0.001/part.tbl.gz", |part| {
-        format!(
```
alamb, line 85: these formatting strings duplicate what is in the Display implementation, so I changed it to use the Display impl

tpchgen/src/generators.rs
```diff
@@ -511,18 +534,26 @@ impl<'a> Iterator for PartGeneratorIterator<'a> {
...
 #[derive(Debug, Clone, PartialEq)]
 pub struct Supplier {
     /// Primary key
     pub s_suppkey: i64,
-    /// Supplier name
-    pub s_name: String,
+    /// Supplier name. Formatted as "Supplier#<n>"
+    pub s_name: i64,
     /// Supplier address
     pub s_address: String,
```
alamb, line 552: There are still some string copies such as this one -- it is due to the fact that the lifetime of the underlying string is tied to the generator (there is a buffer) so I can't return it without the item referring to the iterator

I think what we need to do is postpone the display here as well. I have an idea but I also don't think it will make much of a difference as the Strings are only used now for the much smaller tables.

tpchgen/src/generators.rs
```diff
@@ -1229,9 +1284,9 @@ pub struct Order<'a> {
     /// Order date
     pub o_orderdate: TPCHDate,
     /// Order priority
-    pub o_orderpriority: String,
-    /// Clerk who processed the order
-    pub o_clerk: String,
+    pub o_orderpriority: &'a str,
```
alamb, line 1287: I suspect these two changes are responsible for most of the performance improvement

alamb:
> For example, instead of having Display for Part be responsible for prefixing the "Manufacturer#" string, what if we had a wrapper type that handled that logic, e.g.:



I think it is a  great idea!

Perhaps we can do it as a follow on PR. 

I did something similar here:
- https://github.com/clflushopt/tpchgen-rs/pull/39

There is also something similar for phone numbeers

tpchgen/src/generators.rs
```diff
@@ -266,17 +281,57 @@ impl<'a> Iterator for RegionGeneratorIterator<'a> {
     }
 }
 
+/// A Part Manufacturer, formatted as "Manufacturer#<n>"
```
alamb, line 284: this is an example of what I think @scsmithr  was suggesting in https://github.com/clflushopt/tpchgen-rs/pull/37#issuecomment-2736851885

Now the display logic is in a new type wrapper so that the consumer of a `PartManufacturer` can decide when/where to format the field

tpchgen/src/generators.rs
alamb: Update: I have another PR to avoid these copies:
- https://github.com/clflushopt/tpchgen-rs/pull/39

tpchgen/src/generators.rs
```diff
@@ -788,31 +819,39 @@ impl Iterator for SupplierGeneratorIterator<'_> {
...
+    pub c_comment: &'a str,
 }
 
-impl fmt::Display for Customer {
+impl fmt::Display for Customer<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "{}|{}|{}|{}|{}|{:.2}|{}|{}|",
+            "{}|Customer#{:09}|{}|{}|{}|{:.2}|{}|{}|",
```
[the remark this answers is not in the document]
clflushopt: This is a great idea, I think having those in the Display impl makes much more sense since we don't have to have them as strings anywhere else.

tpchgen-cli/src/main.rs
clflushopt: :brain:

alamb:
I merged up from main to resolve the conflicts in this PR

tpchgen/src/generators.rs
```diff
@@ -144,6 +152,13 @@ impl<'a> Iterator for NationGeneratorIterator<'a> {
 }
 
 /// The REGION table
+/// The Display trait is implemented to format the line item data as a string
```
clflushopt, line 155: To keep the same "pattern" with previous comment can we have a new line in between the top level comment and the next one.
```suggestion

/// The Display trait is implemented to format the line item data as a string
```

clflushopt (approved):
LGTM nothing blocking feel free to resolve conversations and merge ! Thanks

alamb:
> I think @scsmithr might be on to something here, because I realized while at lunch today that the approach of pushing down everything to `Display` makes more sense if we want to leverage columnar generation here. Since every column has already it's own generator we can push down that logic down to `Display` and then have a parametric way to "advance rows" across generators which is the only tricky part for doing column oriented generation here I think.

100% agree

In terms of column oriented generation, what I am thinking is getting to the point we can do it like this (this is using arrow) and likely the approach I will try when creating parquet
- #46 

```rust
// Buffer 1000 rows (this requires that the LineItems are &'static lifetime)
let line_items = linegenerator.take(1000).collect::Vec<_>();

// now, form colums one by one:

// create l_orderkey
let l_orderkey: PrimitiveArray<Int64Type> = line_items.iter().map(|line_item| line_item.l_orderkey).collect();
...
// collect things that are &str directly
let l_returnflag: StringViewArray = line_items.iter().map(|line_item| line_item.l_return_flag).collect();
// collect things that are Display like using builder to avoid allocations
let mut o_clerk_builder = StringArrayBuilder::with_capacity(10000);
line_items.iter().for_each(|line_item| write!(o_clerk_builder, "{}", line_item.o_clerk).unwrap())
let o_clerk = o_clerk_builder.finish()
```

alamb:
I am about out of time for the night so I plan to merge this in once the CI tests pass

alamb:
🚀

PR #38, opened 2025-03-19
Make `'static` lifetime explicit in Generators

alamb:
It wasn't clear to me that the lifetimes on the Generators introduced by @scsmithr  in the following PR is `'static`
- https://github.com/clflushopt/tpchgen-rs/pull/35

Thus I would like to make this more explicit by adding some comments

### DIFF
--- a/tpchgen/src/generators.rs
+++ b/tpchgen/src/generators.rs
@@ -28,8 +28,10 @@ impl Default for NationGenerator<'_> {
 impl<'a> NationGenerator<'a> {
     /// Creates a new NationGenerator with default distributions and text pool
     ///
-    /// Note that the returned structure's lifetime is `&'static`
-    pub fn new() -> Self {
+    /// The generator's lifetime is `&'static` because it references global
+    /// [`Distribution]`s and thus can be shared safely between threads.
+    pub fn new() -> NationGenerator<'static> {
+        // Note: use explicit lifetime to ensure this remains `&'static`
         Self::new_with_distributions_and_text_pool(
             Distributions::static_default(),
             TextPool::get_or_init_default(),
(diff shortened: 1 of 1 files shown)


tpchgen/src/generators.rs
```diff
@@ -26,6 +26,8 @@ impl Default for NationGenerator<'_> {
 
 impl<'a> NationGenerator<'a> {
     /// Creates a new NationGenerator with default distributions and text pool
+    ///
+    /// Note that the returned structure's lifetime is `&'static`
     pub fn new() -> Self {
```
alamb, line 31: I also considered changing the signature of the generators to look like the following, which I somewhat prefer as it is more explicit, but it is also substantially more verbose and may constrain future work

```suggestion
    pub fn new() -> NationGenerator<'static> {
```

However, I figured comments are non objectionable. 

I am happy to make the code change too if you prefer

scsmithr: I think a signature would be reasonable, but it'd need to be placed in its own impl block:

```rust
impl Default for NationGenerator<'static> {
    fn default() -> Self {
        Self::new()
    }
}

impl NationGenerator<'static> {
    /// Creates a new NationGenerator with default distributions and text pool
    pub fn new() -> NationGenerator<'static> {
        Self::new_with_distributions_and_text_pool(
            Distributions::default(),
            TextPool::get_or_init_default(),
        )
    }
}
```

alamb: let's see what @clflushopt thinks

clflushopt: Having it's own `impl` block does make it nicer I would say, my only subjective opinion is that if I were a library user I'd like the clarity it offers. Otherwise LGTM feel free to merge with whichever makes you both comfortable here.

clflushopt (approved):
LGTM

tpchgen/src/generators.rs
alamb: I'll make the change, but I won't have a chance to do so until tomorrow

clflushopt: NP

alamb: I added explicit static lifetimes in 97bc98cce2cad986ebf27a12e7909c337f66944f

clflushopt: Thanks ! Will merge !

PR #40, opened 2025-03-19
Remove lazy_static dep, replace with std::sync::LazyLock

scsmithr:
Removes an extra dep, using stdlib's version.


scsmithr:
cc @alamb @clflushopt

alamb (approved):
Looks good to me -- thanks @scsmithr

Issue #42, opened 2025-03-19
[FEATURE] Generators should accept distribution references


scsmithr:
**Is your feature request related to a problem? Please describe.**

Currently they all call `Distributions::default`. Since distributions are readonly, we can pass in references instead.

Similar to `TextPool`, we can have a default static for `Distributions` that can be used (which would ensure generator lifetimes remain `'static` when `new` is used).

**Describe the solution you'd like**

Something like this:

```rust
/// Generator for Nation table data
pub struct NationGenerator<'a> {
    distributions: &'a Distributions,
    text_pool: &'a TextPool,
}

impl Default for NationGenerator<'static> {
    fn default() -> Self {
        Self::new()
    }
}

impl<'a> NationGenerator<'a> {
    /// Creates a new NationGenerator with default distributions and text pool
    pub fn new() -> Self {
        Self::new_with_distributions_and_text_pool(
            Distributions::get_or_init_default(), // Use global static by default.
            TextPool::get_or_init_default(),
        )
    }

    /// Creates a NationGenerator with the specified distributions and text pool
    pub fn new_with_distributions_and_text_pool(
        distributions: &'a Distributions,
        text_pool: &'a TextPool,
    ) -> Self {
        NationGenerator {
            distributions,
            text_pool,
        }
    }

    /// Returns an iterator over the nation rows
    pub fn iter(&self) -> NationGeneratorIterator {
        NationGeneratorIterator::new(self.distributions.nations(), self.text_pool)
    }
}
```

**Describe alternatives you've considered**

**Additional context**

scsmithr:
@clflushopt didn't mean to assign you. I can go ahead a make this change real quick.

PR #43, opened 2025-03-19
Pass distribution references to generators

scsmithr:
Closes https://github.com/clflushopt/tpchgen-rs/issues/42

Passes references to generators instead of each creating the default one.

The goal of this change is to remove needing to hang onto the generator after creating the iterator (since it owned the distribution data).

Retains the `'static` semantics for all `new` methods on the generators.

### DIFF
--- a/tpchgen/src/distribution.rs
+++ b/tpchgen/src/distribution.rs
@@ -1,5 +1,8 @@
 use crate::random::RowRandomInt;
-use std::io::{self, BufRead};
+use std::{
+    io::{self, BufRead},
+    sync::LazyLock,
+};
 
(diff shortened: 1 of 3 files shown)


scsmithr:
cc @clflushopt @alamb

alamb (approved):
Looks good to me

scsmithr:
First commit didn't _completely_ fix my woes.

Second commit properly ensures that the iterators we get back are not tied to the lifetime of the generator that create it. Essentially this just says that iterators only depend on the textpool and/or distributions it's using.

clflushopt (approved):
LGTM thanks @scsmithr

Issue #44, opened 2025-03-19
[FEATURE] Preserve i32 for "decimal" fields


scsmithr:
**Is your feature request related to a problem? Please describe.**

E.g. `l_discount` is stored as an f64. It should instead be stored as an exact decimal instead of computing it as `discount as f64 / 100.0`.

**Describe the solution you'd like**

Add a `TpchDecimal` type that wraps an `i32` and represents a `Decimal(15,2)`. This should be used for all fields that are considered to be decimals according to the tpch spec.

The `Display` implementation should format it as appropriate.

An important thing is the original `i32` should be accessible to allow converting to exact decimal values for systems that can represent decimals.

An `as_f64` method on this type might be useful if exact decimal semantics aren't needed.

**Describe alternatives you've considered**

**Additional context**

clflushopt:
@scsmithr thanks for opening the issue; I agree I think it makes sense to keep it as such.

Issue #5, opened 2025-03-10
Rust-ify the code


clflushopt:
The code is currently a very verbose Java port is low quality, the goal is to have something more idiomatic.

- Try and maximize the usage of `const fn` evaluation.
- Use a single trait with static dispatch to encapsulate random number generation.
- Use macros to cut down some repetitive code.
- Refactor the implementation into actual modules `tables` for the record schemas, split `dates.rs` contents.
- Refactor the API to allow passing a `Sink<T: Partquet + Arrow>` so we can allow users to pipe the generators to their custom sinks.
- Drop `clone`s and `unwrap`s.
- Improve error handling and stop panic-ing with expects.

clflushopt:
I have some small changes in mind here, notably improving the overall generators API, using the type system more effectively and dropping all Java-isms I inherited from the porting process; but will keep it small for now until we close #6 and #10 since those are higher priority in my opinion.

alamb:
In my mind I think the current codebase is quite readable.

These three I am not sure about 

> Try and maximize the usage of const fn evaluation.
> Use a single trait with static dispatch to encapsulate random number generation.
> Use macros to cut down some repetitive code.

These ones sound good to me:

> Refactor the implementation into actual modules tables for the record schemas, split dates.rs contents.
> Refactor the API to allow passing a Sink<T: Partquet + Arrow> so we can allow users to pipe the generators to their custom sinks.
> Drop clones and unwraps.
> Improve error handling and stop panic-ing with expects.


BTW the first thing I thing we could do is start running `cargo fmt` and `cargo clippy` as part of CI

clflushopt:
I'll add `cargo fmt + clippy` as part of the workflow.

alamb:
Sounds good to me

PR #48, opened 2025-03-20
Add decimal type, avoid floating point calculations (7% faster)

scsmithr:
- Closes https://github.com/clflushopt/tpchgen-rs/issues/44

Opted to go for i64 for the internal representation since a few of the existing float were being created from computations on table keys (i64).

### DIFF
--- a/tpchgen/src/generators.rs
+++ b/tpchgen/src/generators.rs
@@ -1,7 +1,7 @@
 use core::fmt;
 
 use crate::dates;
-use crate::decimal::Decimal;
+use crate::decimal::TPCHDecimal;
 use crate::distribution::Distribution;
 use crate::distribution::Distributions;
 use crate::random::RandomPhoneNumber;
@@ -345,7 +345,7 @@ pub struct Part<'a> {
     /// Part container
     pub p_container: &'a str,
     /// Part retail price
-    pub p_retailprice: Decimal,
+    pub p_retailprice: TPCHDecimal,
     /// Variable length comment
     pub p_comment: &'a str,
 }
@@ -532,7 +532,7 @@ impl<'a> PartGeneratorIterator<'a> {
             p_type: self.type_random.next_value(),
             p_size: self.size_random.next_value(),
             p_container: self.container_random.next_value(),
-            p_retailprice: Decimal(Self::calculate_part_price(part_key)),
+            p_retailprice: TPCHDecimal(Self::calculate_part_price(part_key)),
             p_comment: self.comment_random.next_value(),
         }
     }
@@ -612,7 +612,7 @@ pub struct Supplier {
     /// Supplier phone number
     pub s_phone: PhoneNumberInstance,
     /// Supplier account balance
-    pub s_acctbal: Decimal,
+    pub s_acctbal: TPCHDecimal,
     /// Variable length comment
     pub s_comment: String,
 }
@@ -845,7 +845,7 @@ impl<'a> SupplierGeneratorIterator<'a> {
             s_address: self.address_random.next_value(),
             s_nationkey: nation_key,
             s_phone: self.phone_random.next_value(nation_key),
-            s_acctbal: Decimal(self.account_balance_random.next_value() as i64),
+            s_acctbal: TPCHDecimal(self.account_balance_random.next_value() as i64),
             s_comment: comment,
         }
     }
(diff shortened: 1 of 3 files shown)


tpchgen/src/generators.rs
```diff
@@ -1723,13 +1724,14 @@ pub struct LineItem<'a> {
     /// Line item number within order
     pub l_linenumber: i32,
     /// Quantity ordered
+    // TODO: Spec has this as decimal.
     pub l_quantity: i64,
```
scsmithr, line 1728: `l_quantity` should be a decimal according to the spec, but switching it over broke some of the current tests against well-known output.

I didn't dig in too deeply, but I'll investigate at some point why that happens. But a quick poke at the trino source showed that they use long and don't format it like they do for the monetary fields, so figured that's probably baked into the output that's produced.

clflushopt (approved):
LGTM one minor nitpick can we name it `TPCHDecimal` this way it's clear to the users that this is a tpch type, similarly to what @alamb did in #27 with  the custom date type as `TPCHDate` see https://github.com/clflushopt/tpchgen-rs/blob/main/tpchgen/src/dates.rs#L57. In the future maybe in #5 I'll link a PR to group the custom types in a `tpch` module.

tpchgen/src/generators.rs
alamb: FWIW the expected output I think also matches the original tpch dbgen program (as in I don't think this is a trino specific behavior). 

Read more about it here:
https://github.com/clflushopt/tpchgen-rs/tree/main/tpchgen/data#comparing-with-other-tpch-dbgen-programs

I also double checked it against the [original dbgen](https://github.com/electrum/tpch-dbgen) program I built from source this morning:

tpchgen/src/generators.rs
```diff
@@ -844,7 +845,7 @@ impl<'a> SupplierGeneratorIterator<'a> {
             s_address: self.address_random.next_value(),
             s_nationkey: nation_key,
             s_phone: self.phone_random.next_value(nation_key),
-            s_acctbal: self.account_balance_random.next_value() as f64 / 100.0,
+            s_acctbal: Decimal(self.account_balance_random.next_value() as i64),
```
alamb, line 848: I do wonder if using an i32 as the internal representation would be more efficient (though I see total price is an i64)

alamb (approved):
Looks good to me -- thank you very much @scsmithr 

I agree with @clflushopt  that it would be better if named this `TPCHDecimal` or something, but I don't feel strongly (and we could do that as a follow on PR too)

I also checked performance,  and this PR seems to make a meaningful performance difference(7%) 👍 

| branch | time |
|--------|--------|
| main | 0m6.213s |
| this pr | 0m5.806s | 

<details><summary>Details</summary>
<p>


Main
```shell
(venv) andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
Generation complete!

real	0m6.213s
user	0m5.939s
sys	0m0.228s
(venv) andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
Generation complete!

real	0m6.202s
user	0m5.897s
sys	0m0.239s
```

This pr

```
(venv) andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
Generation complete!

real	0m5.806s
user	0m5.544s
sys	0m0.243s
(venv) andrewlamb@Andrews-MacBook-Pro-2:~/Software/tpchgen-rs$ time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
Generation complete!

real	0m5.874s
user	0m5.574s
sys	0m0.273s
```

</p>
</details>

tpchgen/src/generators.rs
scsmithr: Probably would be, but figured if there's a difference, it's likely tiny. I was originally writing this with i32, but decided to switch to i64 to reduce the likelihood of me introducing a bug. It's easier to cast everything to i64, and the i64 -> i32 casts are just from well-defined min/max bounds that fit in i32.

alamb:
Thanks @scsmithr

Issue #46, opened 2025-03-19
[FEATURE] Directly write parquet


alamb:
**Is your feature request related to a problem? Please describe.**
I would like the ability to make the TPCH dataset as parquet files

It will be possible to do this by creating `.tbl` files and then using some other tool (like `datafusion-cli` to covert the data). This has at least two downsides:
1. It is slow as you have to create a text format (`.tbl`) and then parse it again
2. It is a poor UX as it requires multiple toops

**Describe the solution you'd like**
However, I would like to be able to create parquet directly. 

**Describe alternatives you've considered**

I suggest
1. adding a `parquet` optional feature flag
2. Adding a `--format` option to the CLI that controls the format

So something like
```shell
tpchgen-cli --format parquet -s 100 --output-dir=.
```

**Additional context**
It might also be nice to create a real CSV (with header, etc) option as well
- https://github.com/clflushopt/tpchgen-rs/issues/53

clflushopt:
Will add it to the least of things for me to tackle ! Thanks for opening the issue, speaking of which do we want to have a facility to generate arrow buffers instead of our custom structs at the library level ?

alamb:
Copying some thoughts from here: https://github.com/clflushopt/tpchgen-rs/pull/37#issuecomment-2738354002

> speaking of which do we want to have a facility to generate arrow buffers instead of our custom structs at the library level ?

I think we will basically need to write code to create arrow arrays to write to parquet efficiently

here is how I imagine the code looking to create record batches to feed:

```rust
// Buffer 1000 rows (this requires that the LineItems are &'static lifetime)
let line_items = linegenerator.take(1000).collect::Vec<_>();

// now, form colums one by one:

// create l_orderkey
let l_orderkey: PrimitiveArray<Int64Type> = line_items.iter().map(|line_item| line_item.l_orderkey).collect();
...
// collect things that are &str directly
let l_returnflag: StringViewArray = line_items.iter().map(|line_item| line_item.l_return_flag).collect();
// collect things that are Display like using builder to avoid allocations
let mut o_clerk_builder = StringArrayBuilder::with_capacity(10000);
line_items.iter().for_each(|line_item| write!(o_clerk_builder, "{}", line_item.o_clerk).unwrap())
let o_clerk = o_clerk_builder.finish()
```

alamb:
BTW here is the TPCH schema used by datafusion

[...]

PR #49, opened 2025-03-20
Add `ARCHITECTURE.md` documentation with design goals (keep data generator crate dependencies minimal)

alamb:
# Rationale
@scsmithr @clflushopt and I were discussing the crates design here: https://github.com/clflushopt/tpchgen-rs/issues/46#issuecomment-2738433070

I wanted to capture the consensus in documentation. 

Please let me know if I got it right

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -20,4 +20,4 @@ has many more dependencies.
 Speed is a very important aspect of this project, and care has been taken to keep 
 the code as fast as possible, using some of the following techniques:
 1. Avoiding heap allocations during data generation
-2. Integer arithmetic instead of floating point arithmetic
+2. Integer arithmetic and display instead of floating point arithmetic and display
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,23 @@
+# Architecture Guide
+
+## Crate Organization
+The project is organized into two crates:
+
+1. `tpchgen`: The core library that implements the data generation logic for TPCH.
+2. `tpchgen-cli`: A CLI tool that uses the `tpchgen` library to generate TPCH data.
+
+## Dependencies
+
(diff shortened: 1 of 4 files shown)


ARCHITECTURE.md
```diff
@@ -0,0 +1,23 @@
...
+
+The `tpchgen-cli` crate is designed to include many useful features, and thus
+has many more dependencies.
+
+## Speed
+
+Speed is a very important aspect of this project, and care has been taken to keep 
+the code as fast as possible, using some of the following techniques:
+1. Avoiding heap allocations during data generation
+2. Integer arithmetic instead of floating point arithmetic
```
scsmithr, line 23: More of a technicality, but I think the speed up with the decimal stuff was from removing float formatting, and less around fp arithmetic.

I also think correctness trumps speed here, and the move to decimal was going for a more correct approach.

scsmithr (approved):
lgtm

ARCHITECTURE.md
alamb: I agree that the motivation for switching to decimal might not have been speed, but I thought it was a good example of things we have done to make this faster. How about this

```suggestion
2. Integer arithmetic and display instead of floating point arithmetic and display
```

scsmithr: Works for me.

tpchgen/data/README.md
```diff
@@ -83,7 +83,8 @@ C data generator to verify they are the same. To do so:
 Step 1: create `tbl` files.
 
 One way to do this is using a docker container that has the classic
-data generator prebuilt, though you could also build it from scratch:
+data generator prebuilt, though you could also build it from 
+[source](https://github.com/electrum/tpch-dbgen):
```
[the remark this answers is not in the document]
clflushopt: lgtm

clflushopt (approved):

PR #50, opened 2025-03-20
Precompute date strings (10% faster)

alamb:
- Part of #6 

While looking around the code, I noticed that significant time was being spent formatting dates, but that there are only ~2500 dates in total. Thus pre-computing the string representation is more efficient

# Changes:
1. Precompute date strings rather than Chrono dates
2. Add to_chrono conversion method
3. Add some tests

# Performance

```shell
time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
```

| branch | time |
|--------|--------|
| main | 0m6.325s |
| this branch | 0m5.667s |


alamb:
Combined with @scsmithr 's PR in https://github.com/clflushopt/tpchgen-rs/pull/50 

We are getting pretty close to 5s for scale factor 1

```
0m5.286s
```

scsmithr (commented):
What's the path to getting a raw date value and not just the string representation (e.g. for converting date values to an arrow date array)?

scsmithr:
> What's the path to getting a raw date value and not just the string representation (e.g. for converting date values to an arrow date array)?

Actually disregard, I missed the `to_chrono` method which works well enough for now.

This looks good to me.

alamb:
> > What's the path to getting a raw date value and not just the string representation (e.g. for converting date values to an arrow date array)?
> 
> Actually disregard, I missed the `to_chrono` method which works well enough for now.
> 
> This looks good to me.

I'll can also and implement a `into_inner()` type method to get the actual inner i32. We should probably do that for all the wrappers actually 🤔

clflushopt (approved):
lgtm

clflushopt:
Closes #6 thanks @alamb !

PR #52, opened 2025-03-21
Add `into_inner` for wrapper types, and derive some more traits`

alamb:
Inspired by @scsmithr's comments https://github.com/clflushopt/tpchgen-rs/pull/50#pullrequestreview-2702915563

We now have date and decimal so called "newtype" wrappers around `i32`/`i64` that allow formatting into a single location

Right now it is easy to go from `i32` -> newtype, but not the other way around

Typically, there is a function in rust programs `into_inner` that does the conversion.

We could also consider adding `From` impls, but I think these functions are more explicit and we can add from impls later

I also added some more `derive` traits while I was at it

### DIFF
--- a/tpchgen/src/dates.rs
+++ b/tpchgen/src/dates.rs
@@ -57,7 +57,7 @@ impl GenerateUtils {
 /// Represents a date (day/year)
 ///
 /// Example display: 1992-01-01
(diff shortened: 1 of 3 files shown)


scsmithr (approved):
lgtm

alamb:
Thanks for the review @scsmithr

Issue #53, opened 2025-03-21
[FEATURE] Directly write csv format (in addition to tbl format)


alamb:
- Similar to https://github.com/clflushopt/tpchgen-rs/issues/46

**Is your feature request related to a problem? Please describe.**
I would like the ability to make the TPCH dataset as csv files (with a header)

The output of tpchdbgen is `.tbl` files which are like csv but
1. have a different separator (`|`)
2. have no header

This makes it hard to use them with other tools 

**Describe the solution you'd like**
However, I would like to be able to create csv files with 

**Describe alternatives you've considered**

I suggest
1. Adding a `--format` option to the CLI that controls the format

So something like
```shell
tpchgen-cli --format csv -s 1 --output-dir=.
```

**Additional context**
- Similar to https://github.com/clflushopt/tpchgen-rs/issues/46

alamb:
I am working on this (as a step towards parallell generation)

PR #55, opened 2025-03-21
Add docs, improve doc tests

alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/53

While working on  I wanted a place to document / test how to write CSV, I got carried away with some introductory text / examples in the docs, which also lead to realize `cargo doc` had a bunch of errors too

Thus here is a PR to:
1. Add some documentation to the crate
2. Add `cargo doc` to the CI tests
3. Break the CI checks into smaller jobs that can run in parallel (to improve cycle time)
4. Gate CI checks so they don't run unless cargo fmt/ clippy / build fail


.github/workflows/conformance.yml
```diff
@@ -1,40 +0,0 @@
-name: TPC-H Conformance Tests
```
alamb, line 1: consolidated into rust.yml so I could gate it on fmt/clippy passing

.github/workflows/rust.yml
```diff
@@ -5,21 +5,91 @@ on:
...
-
     steps:
       - uses: actions/checkout@v4
       - name: Run cargo-fmt
         run: cargo fmt --all -- --check
       - name: Run cargo-clippy
         run: cargo clippy -- -D warnings
-      - name: Build
-        run: cargo build --verbose
+      - name: Check
```
alamb, line 22: I broke the jobs into multiple ones that ran in parallel once `clippy` and `fmt` and `check` pass so avoid wasting I runner time on jobs that will not compile

.github/workflows/rust.yml
```diff
@@ -5,21 +5,91 @@ on:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name : Build docs
+        env:
+          RUSTDOCFLAGS: "-D warnings"
+        run: cargo doc --no-deps --workspace
+
+
+  # Conformance tests (compare to existing  implementation)
+  conformance-test:
```
alamb, line 65: Since this one takes a bit I gated it on the simpler unit tests passing so we don't waste builder time on the expensive one if the CI will fail on cheaper tests.

alamb:
@clflushopt  and @scsmithr  - I am going to merge this in as I think it should be non controversial: it adds documentation, fixes some warnings during documentation build, and make the CI more efficient

I don't want to end up with stacked PRs for review

I am happy to address any concerns or comments as follow on PRs

alamb:
YOLO

PR #54, opened 2025-03-21
Implement output as CSV

alamb:
Add ability to print tpch tables as csv files

- closes https://github.com/clflushopt/tpchgen-rs/issues/53

This is a step towards parallel generation and parquet output

I also verified you can query these files from datafusion-cli. A teaser:
```sql
> select * from '/tmp/tpchdbgen-rs/lineitem.csv' limit 10;
+------------+-----------+-----------+--------------+------------+-----------------+------------+-------+--------------+--------------+------------+--------------+---------------+-------------------+------------+-------------------------------------+
| l_orderkey | l_partkey | l_suppkey | l_linenumber | l_quantity | l_extendedprice | l_discount | l_tax | l_returnflag | l_linestatus | l_shipdate | l_commitdate | l_receiptdate | l_shipinstruct    | l_shipmode | l_comment                           |
+------------+-----------+-----------+--------------+------------+-----------------+------------+-------+--------------+--------------+------------+--------------+---------------+-------------------+------------+-------------------------------------+
| 1          | 15519     | 785       | 1            | 17         | 24386.67        | 0.04       | 0.02  | N            | O            | 1996-03-13 | 1996-02-12   | 1996-03-22    | DELIVER IN PERSON | TRUCK      | egular courts above the             |
| 1          | 6731      | 732       | 2            | 36         | 58958.28        | 0.09       | 0.06  | N            | O            | 1996-04-12 | 1996-02-28   | 1996-04-20    | TAKE BACK RETURN  | MAIL       | ly final dependencies: slyly bold   |
| 1          | 6370      | 371       | 3            | 8          | 10210.96        | 0.1        | 0.02  | N            | O            | 1996-01-29 | 1996-03-05   | 1996-01-31    | TAKE BACK RETURN  | REG AIR    | riously. regular, express dep       |
| 1          | 214       | 465       | 4            | 28         | 31197.88        | 0.09       | 0.06  | N            | O            | 1996-04-21 | 1996-03-30   | 1996-05-16    | NONE              | AIR        | lites. fluffily even de             |
| 1          | 2403      | 160       | 5            | 24         | 31329.6         | 0.1        | 0.04  | N            | O            | 1996-03-30 | 1996-03-14   | 1996-04-01    | NONE              | FOB        |  pending foxes. slyly re            |
[...]

PR #58, opened 2025-03-21
Parallel output in tpchgen-cli (Nx faster, where N is number of cores)

alamb:
- closes https://github.com/clflushopt/tpchgen-rs/issues/10
- closes https://github.com/clflushopt/tpchgen-rs/pull/34
- builds on https://github.com/clflushopt/tpchgen-rs/pull/54

Features:
1. Logging support via `-v` and `RUST_LOG`
2. Parallel generation for all tables
3. No new dependencies on the tpch core crate and minor changes

# Performance 🌶️ 
In single threaded mode, the generator can create 250MB/sec with each core which is 4GB/s on my 16 core laptop

This scales linearly with 

Here are performance measurements for SF=100 on my Mac M3 (with 16 cores)

| branch | speed | speed (writing to /dev/null) |
|--------|--------|--------|
| main | 6m1.984s | N/A
| this branch | 1m0.099s | 0m28.732s | 

```shell
time target/release/tpchgen-cli -s 100 --output-dir=/tmp/tpchdbgen-rs
```

It actually turns out this can fully saturate the disk bandwidth on my Mac M3 laptop which caps out at 2GB/sec

If I hard code the generator to just throw the data away rather than writing, it keeps all the cores busy and writes at a blistering 4GB/sec

### DIFF
--- a/tpchgen-cli/src/generate.rs
+++ b/tpchgen-cli/src/generate.rs
@@ -101,7 +101,7 @@ where
 
 /// A simple buffer recycler to avoid allocating new buffers for each part
 ///
-/// Clones share the same undrlying recycler, so it is not thread safe
+/// Clones share the same underlying recycler, so it is not thread safe
 #[derive(Debug, Clone)]
 struct BufferRecycler {
     buffers: Arc<Mutex<VecDeque<Vec<u8>>>>,
--- a/tpchgen-cli/src/generate.rs
+++ b/tpchgen-cli/src/generate.rs
@@ -26,7 +26,8 @@ pub trait Sink: Send {
     /// Write all data from the buffer to the sink
     fn sink(&mut self, buffer: &[u8]) -> Result<(), io::Error>;
 
-    fn finish(self) -> Result<(), io::Error>;
+    /// Complete and flush any remaining data from the sink
+    fn flush(self) -> Result<(), io::Error>;
 }
 
 /// Creates data from a set of [`Source`] in parallel sending data in order to the Sink.
@@ -84,7 +85,7 @@ where
             sink.sink(&buffer)?;
             captured_recycler.return_buffer(buffer);
         }
-        sink.finish()
+        sink.flush()
     });
 
     // drive the stream to completion
--- a/tpchgen-cli/src/generate.rs
+++ b/tpchgen-cli/src/generate.rs
@@ -1,0 +1,134 @@
+//! Parallel data generation: [`Source`] and [`Sink`] and [`generate_in_chunks`]
+//!
+//! These traits and function are used to generate data in parallel and write it to a sink
+//! in streaming fashion (chunks). This is useful for generating large datasets that don't fit in memory.
+
+use futures::StreamExt;
+use log::debug;
+use std::collections::VecDeque;
+use std::io;
+use std::sync::{Arc, Mutex};
+use tokio::task::JoinSet;
+
+/// Something that knows how to generate data into a buffer
+///
+/// For example, this is implemented for the different generators in the tpchgen
+/// crate
+pub trait Source: Send {
+    /// generates the data for this generator into the buffer, returning the buffer.
+    fn create(self, buffer: Vec<u8>) -> Vec<u8>;
+}
+
+/// Something that can write the contents of a buffer somewhere
+///
+/// For example, this is implemented for a file writer
+pub trait Sink: Send {
+    /// Write all data from the buffer to the sink
+    fn sink(&mut self, buffer: &[u8]) -> Result<(), io::Error>;
+
+    fn finish(self) -> Result<(), io::Error>;
+}
+
+/// Creates data from a set of [`Source`] in parallel sending data in order to the Sink.
+///
+///
+/// G: Generator
+/// I: Iterator<Item = G>
+/// S: Sink that writes buffers somewhere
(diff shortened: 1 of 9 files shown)


alamb:
Ok, I need to stop for today, but this is looking promising

tpchgen-cli/src/generate.rs
```diff
@@ -0,0 +1,128 @@
+//! Parallel data generation: [`Source`] and [`Sink`] and [`generate_in_chunks`]
```
alamb, line 1: All the parallel generation code is in this module

tpchgen-cli/src/main.rs
```diff
@@ -10,22 +10,42 @@
 //! -T, --tables     Tables to generate data for
 //! -F, --format     Output format for the data (CSV or Parquet)
 //! -O, --output     Output directory for the generated data
-//
-// The main function is the entry point for the CLI and it uses the `clap` crate
-// to parse the command line arguments and then generate the data.
-
-// tpchgen-cli/src/main.rs
+//! -v, --verbose    Verbose output
```
alamb, line 13: new logging feature

tpchgen-cli/src/main.rs
```diff
@@ -90,14 +120,53 @@ enum OutputFormat {
...
-fn main() -> io::Result<()> {
+#[tokio::main]
+async fn main() -> io::Result<()> {
     // Parse command line arguments
     let cli = Cli::parse();
-    cli.main()
+    cli.main().await
+}
+
+/// macro to create a Cli function for generating a table
```
alamb, line 130: avoided a lot of repetition with macros

tpchgen/src/generators.rs
```diff
@@ -23,16 +23,21 @@ pub struct NationGenerator<'a> {
...
     /// Creates a new NationGenerator with default distributions and text pool
     ///
+    /// Nations does not depend on the scale factor or the part number. The signature of
+    /// this method is provided to be consistent with the other generators, but the
+    /// parameters are ignored.
+    ///
     /// The generator's lifetime is `&'static` because it references global
     /// [`Distribution]`s and thus can be shared safely between threads.
-    pub fn new() -> NationGenerator<'static> {
+    pub fn new(_scale_factor: f64, _part: i32, _part_count: i32) -> NationGenerator<'static> {
```
alamb, line 40: I made these methods take the same arguments as the other generators to simplify the macros (all generators now have the same signature)

tpchgen/src/generators.rs
```diff
@@ -433,6 +443,11 @@ impl<'a> PartGenerator<'a> {
         }
     }
 
+    /// Return the row count for the given scale factor and generator part count
```
alamb, line 446: I refactored out the row count calculation into a function that can be called from outside (needed to calculate how many parts for each table)

alamb:
I plan to merge this PR tomorrow unless anyone else would like time to comment on it.

clflushopt:
I will take a look at it tonight ! Thanks @alamb

tpchgen-cli/src/generate.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Something that can write the contents of a buffer somewhere
+///
+/// For example, this is implemented for a file writer
+pub trait Sink: Send {
+    /// Write all data from the buffer to the sink
+    fn sink(&mut self, buffer: &[u8]) -> Result<(), io::Error>;
+
+    fn finish(self) -> Result<(), io::Error>;
```
clflushopt, line 29: question: I have no preference here, but I was wondering if we shouldn't rename `finish` to `flush` to keep it "semantically closed" to how a sink might behave e.g. `flush()` calls `flush` and closes the file or buffered writer underlying.

tpchgen-cli/src/generate.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Creates data from a set of [`Source`] in parallel sending data in order to the Sink.
```
clflushopt, line 32: suggestion: Can you clarify that `Source` here is a distribution or row generator and clarify that order means  order in which generators are specified.

tpchgen-cli/src/generate.rs
```diff
@@ -0,0 +1,134 @@
...
+    drop(tx); // drop last tx reference to stop the writer
+
+    // wait for writer to finish
+    debug!("waiting for writer task to complete");
+    writer_task.await.expect("writer task panicked")
+}
+
+/// A simple buffer recycler to avoid allocating new buffers for each part
+///
+/// Clones share the same undrlying recycler, so it is not thread safe
```
clflushopt, line 104:
```suggestion
/// Clones share the same underlying recycler, so it is not thread safe
```

clflushopt (approved):
LGTM; awesome stuff !!!

tpchgen-cli/src/generate.rs
alamb: Excellent idea -- done in 4985895

tpchgen-cli/src/generate.rs
alamb: Done in b611865

alamb:
BTW with the other improvements on main, when merging it takes less than 5 seconds on my laptop to make TPCH SF 10 🚀 

```shell
time target/release/tpchgen-cli -s 10  --output-dir=/tmp/tpchdbgen-rs

real	0m4.895s
user	0m36.096s
sys	0m2.910s
```

Issue #56, opened 2025-03-21
[FEATURE] Reduce latency to create first row ( initial table generation time)


alamb:
**Is your feature request related to a problem? Please describe.**
Right now, there is a non trivial period of time (approx 2 seconds on release builds) generating the first row of data for any table

This is due to the fact that the text pool Distributions are loaded on first access.

This is inefficient because it happens the first time in any process (and for example, means each doc examples take many seconds as they eash run in their own 

You can see an example of how silly this is by noticing that creating the nations table SF1 (25 rows, 2Kb)  takes almost 2 seconds to create
```shell
$ time target/release/tpchgen-cli -s 1 --tables=nation --output-dir=/tmp/tpchdbgen-rs
Generation complete!

real	0m1.960s
```

While the lineitem table (6.1M rows, 759 MB) only 4.4 seconds
```shell
$ time target/release/tpchgen-cli -s 1 --tables=line-item --output-dir=/tmp/tpchdbgen-rs
Generation complete!

real	0m4.472s
```

```shell
wc /tmp/tpchdbgen-rs/nation.tbl /tmp/tpchdbgen-rs/lineitem.tbl 
      25     276    2224 /tmp/tpchdbgen-rs/nation.tbl
 6001215 35481203 759863287 /tmp/tpchdbgen-rs/lineitem.tbl
 6001240 35481479 759865511 total
```
**Describe the solution you'd like**
It would be really nice if we could reduce the time required to generate the first row. 

**Describe alternatives you've considered**
We could potentially make the actual parsing logic more efficient.

Another alternative is to automatically generate static tables at build time, using a `build.rs` build script

For example, the load script could parse the initial lines and break them up into `&'static [String]` at least to save the time on each process run

**Additional context**

alamb:
FYI @scsmithr  I am not sure if this is an important feature for you (faster initial gen time)

alamb:
I don't plan to spend much more time on this until after csv / parallel gen / parquet

scsmithr:
I can whip something up for removing the indexmap and see if that helps.

alamb:
https://github.com/clflushopt/tpchgen-rs/pull/62 cut the time by half (thanks @scsmithr !)

I profiled again 
```shell
time target/release/tpchgen-cli --tables line-item -s 0.001 --output-dir=/tmp/tpchdbgen-rs
```

That takes about a second. Almost all the time is spent copying memory around

[...]

PR #59, opened 2025-03-21
(POC): Remove some hashmap lookups in distribution lookup

scsmithr:
Partially addresses https://github.com/clflushopt/tpchgen-rs/issues/56

Also removes IndexMap

Wip. The timings are down, but there might be alternative approaches here that I may try to explore.

This approach would also remove some flexibility with what we allow in a distribution, but I don't think that matter for this.

# Timings

```
$ time cargo run --release -- -s 0.01 --output-dir=/tmp/tpchdbgen-rs-001
```

Tiny SF to capture the initialization time, not actual data generation.

`main`: 2.851s, 2.694s, 2.764s
this branch: 1.860s, 1.867s, 1.864s

### DIFF
--- a/tpchgen/Cargo.toml
+++ b/tpchgen/Cargo.toml
@@ -8,7 +8,6 @@ authors = ["clflushopt", "alamb"]
 # See ../ARCHITECTURE.md for more details
 [dependencies]
 chrono = "0.4.40"
-indexmap = "2.7.1"
 
 [dev-dependencies]
 flate2 = "1.1.0"
--- a/tpchgen/src/distribution.rs
+++ b/tpchgen/src/distribution.rs
@@ -215,28 +215,47 @@ impl DistributionLoader {
 /// Static global instance of the default distributions.
 ///
 /// Initialized once on first access.
-static DEFAULT_DISTRIBUTIONS: LazyLock<Distributions> = LazyLock::new(Distributions::default);
+static DEFAULT_DISTRIBUTIONS: LazyLock<Distributions> =
+    LazyLock::new(|| Distributions::try_load_defualt().unwrap());
 
 /// Distributions wraps all TPC-H distributions and provides methods to access them.
 #[derive(Debug, Clone)]
 pub struct Distributions {
+    articles: Distribution,
+    adjectives: Distribution,
+    adverbs: Distribution,
+    nouns: Distribution,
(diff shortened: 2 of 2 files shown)


tpchgen/Cargo.toml
```diff
@@ -8,7 +8,6 @@ authors = ["clflushopt", "alamb"]
 # See ../ARCHITECTURE.md for more details
 [dependencies]
 chrono = "0.4.40"
-indexmap = "2.7.1"
```
alamb, line 11: 👍

tpchgen/src/distribution.rs
```diff
@@ -217,28 +215,47 @@ impl DistributionLoader {
...
-    /// Creates a new distributions wrapper.
-    pub fn new(distributions: IndexMap<String, Distribution>) -> Self {
-        Distributions { distributions }
+        Ok(Distributions {
+            articles: remove_dist("articles")?,
+            adjectives: remove_dist("adjectives")?,
+            adverbs: remove_dist("adverbs")?,
+            nouns: remove_dist("nouns")?,
+            verbs: remove_dist("verbs")?,
+            auxiliaries: remove_dist("auxillaries")?, // P.S: The correct spelling is `auxiliaries` which is what we use.
```
alamb, line 255: 😆

alamb (commented):
Thanks @scsmithr  -- this is looking great. Let me know when / it is ready for a review

scsmithr:
> I started going through this PR -- the idea is really nice. I will try and bash out the conversion for the remaining distributions

Yeah my bad for the slowness here. Feel free to take it over.

My only real concern is we lose on flexibility on what distributions we can use. But that's my programmer brain talking -- in reality we only care about that one distribution file.

alamb:
> Yeah my bad for the slowness here. Feel free to take it over.

No worries -- this is a great step. I also have some ideas about how to avoid hash maps entirely. But I'll do that as a follow on PR potentially. 

Right now me and copilot are going to apply your pattern to all the remaining distributions :)

> My only real concern is we lose on flexibility on what distributions we can use. But that's my programmer brain talking -- in reality we only care about that one distribution file.

Yeah, I agree -- the TPCH data generator hasn't changed for at least 14 years. I don't think flexibility is particularly important

PR #62, opened 2025-03-25
Remove  hashmap lookups in distribution creation, remove IndexMap dependency (2x startup time speedup)

alamb:
- Part of https://github.com/clflushopt/tpchgen-rs/issues/56
- Closes https://github.com/clflushopt/tpchgen-rs/pull/59

This PR is based on great work from @scsmithr  in https://github.com/clflushopt/tpchgen-rs/pull/59 and makes initial Distribution loading time 2x faster. 

The key change is to loading Distributions so the names are resolved once on load rather than on each access. 

This significantly speeds up the startup time

Testing with
```shell
time target/release/tpchgen-cli --tables line-item -s 0.001 --output-dir=/tmp/tpchdbgen-rs
```

| Header | Header |
|--------|--------|
| main | 0m2.036s |
| this branch | 0m0.985s |

### DIFF
--- a/tpchgen/src/distribution.rs
+++ b/tpchgen/src/distribution.rs
@@ -215,28 +215,47 @@ impl DistributionLoader {
 /// Static global instance of the default distributions.
 ///
 /// Initialized once on first access.
-static DEFAULT_DISTRIBUTIONS: LazyLock<Distributions> = LazyLock::new(Distributions::default);
+static DEFAULT_DISTRIBUTIONS: LazyLock<Distributions> =
+    LazyLock::new(|| Distributions::try_load_defualt().unwrap());
 
 /// Distributions wraps all TPC-H distributions and provides methods to access them.
 #[derive(Debug, Clone)]
 pub struct Distributions {
+    articles: Distribution,
+    adjectives: Distribution,
+    adverbs: Distribution,
+    nouns: Distribution,
+    verbs: Distribution,
+    auxiliaries: Distribution,
+
     distributions: HashMap<String, Distribution>,
 }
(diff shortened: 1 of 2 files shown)


tpchgen/src/distribution.rs
```diff
@@ -103,73 +98,20 @@ impl Distribution {
         }
         unreachable!("Cannot get random value from an invalid distribution")
     }
-}
-
-/// DistributionLoader provides functionality to load TPC-H distributions from a text format.
```
alamb, line 108: I moved this code into `Distribution`

tpchgen/src/distribution.rs
```diff
@@ -217,28 +159,118 @@ impl DistributionLoader {
...
-static DEFAULT_DISTRIBUTIONS: LazyLock<Distributions> = LazyLock::new(Distributions::default);
+static DEFAULT_DISTRIBUTIONS: LazyLock<Distributions> =
+    LazyLock::new(|| Distributions::try_load_default().unwrap());
 
 /// Distributions wraps all TPC-H distributions and provides methods to access them.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct Distributions {
-    distributions: IndexMap<String, Distribution>,
+    articles: Distribution,
```
alamb, line 168: They key change is that now the distributions are loaded into the correct field at load time, rather than on access

scsmithr (approved):
🚀 🚀 🚀

alamb:
Thank you for the super quick review @scsmithr

PR #64, opened 2025-03-25
feat: Use lookup tables for julian date format (3-5% improvement)

clflushopt:
This change introduces a lookup table that's built on first access for julian date format used when generating the lineitem table. The optimization eliminates a path that we spend about 14% of total running time in (profile sampled).

Here's a before profile where 14% of the time was spent calling `julian`.
before-julian-date-optim

An after profile with the block entirely gone due to lazily initializing it once and using a lookup table for the rest of the dates.
after-julian-date-optim

The trade-off here is entirely about moving runtime CPU cost to memory which is not a lot (I estimated about 9 KiB from `4 * 2557` where 2557 is the number of `i32`'s we store in the lookup table.

Below are the timings captured from `main` and this branch on an 9950X workstation (single core used)  from a `release` build.

`main` :

```
clflushopt@workstation ~/P/tpchgen-rs (main)> time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
Generation complete!

________________________________________________________
Executed in    3.74 secs    fish           external
   usr time    3.51 secs  201.00 micros    3.51 secs
   sys time    0.22 secs   97.00 micros    0.22 secs

```

This branch:

```
clflushopt@workstation ~/P/tpchgen-rs (cl/feat/julian-date-lookup-tables)> time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs
Generation complete!

________________________________________________________
Executed in    3.60 secs    fish           external
   usr time    3.39 secs    0.00 micros    3.39 secs
   sys time    0.21 secs  332.00 micros    0.21 secs
```

Edit: Had to replace SVG's with snapped pictures but the profile can be replicated using `cargo flamegraph` & `perf` as such `cargo flamegraph -- -s 1 --output-dir=/tmp/tpchdbgen-rs`

### DIFF
--- a/tpchgen/src/dates.rs
+++ b/tpchgen/src/dates.rs
@@ -22,6 +22,14 @@ const MONTH_YEAR_DAY_START: [i32; 13] =
 /// value: strings like 1992-01-01
 static DATE_TO_STRING: LazyLock<Vec<String>> = LazyLock::new(make_date_index);
 
+/// Lookup table for julian date format used to check if a given date is in
+/// the past as part of the lineitem generation.
+static JULIAN_DATE: LazyLock<Vec<i32>> = LazyLock::new(|| {
+    (0..TOTAL_DATE_RANGE)
+        .map(|date| julian(date + MIN_GENERATE_DATE))
+        .collect()
+});
+
 pub struct GenerateUtils;
 
 impl GenerateUtils {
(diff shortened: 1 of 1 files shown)


tpchgen/src/dates.rs
```diff
@@ -95,33 +103,13 @@ impl TPCHDate {
     }
 
     /// Checks if a date is in the past
-    pub const fn is_in_past(date: i32) -> bool {
-        Self::julian(date) <= CURRENT_DATE
-    }
-
-    /// Converts to julian date format
```
alamb, line 102: it is nice to remove this duplication too

PR #67, opened 2025-03-25
Speed up initial TextPool generation more (15% faster)

alamb:
- Part of https://github.com/clflushopt/tpchgen-rs/issues/56

More random things I noticed while trying to make startup faster. I will comment in line

Performance:

```
time target/release/tpchgen-cli -s 0.001 --tables=line-item --output-dir=/tmp/tpchdbgen-rs
```

| branch | time |
|--------|--------|
| main | 0m0.960s |
| this branch | 0m0.833s | 

The biggest change is to avoid an N^2 algorithm in text generation that I noticed while looking at the flamegraph

Screenshot 2025-03-25 at 7 04 48 AM

### DIFF
--- a/tpchgen/src/text.rs
+++ b/tpchgen/src/text.rs
@@ -80,14 +80,14 @@ impl TextPool {
         let syntax = distributions.grammar().random_value(random);
         let max_length = syntax.len();
 
-        for i in (0..max_length).step_by(2) {
-            match syntax.chars().nth(i).unwrap() {
+        for c in syntax.chars().take(max_length).step_by(2) {
+            match c {
                 'V' => Self::generate_verb_phrase(distributions, builder, random),
                 'N' => Self::generate_noun_phrase(distributions, builder, random),
                 'P' => {
                     let preposition = distributions.prepositions().random_value(random);
                     builder.append(preposition);
-                    builder.append(" the ");
+                    builder.append_bytes(b" the ");
                     Self::generate_noun_phrase(distributions, builder, random);
                 }
                 'T' => {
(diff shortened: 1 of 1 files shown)


tpchgen/src/text.rs
```diff
@@ -80,14 +80,14 @@ impl TextPool {
         let syntax = distributions.grammar().random_value(random);
         let max_length = syntax.len();
 
-        for i in (0..max_length).step_by(2) {
-            match syntax.chars().nth(i).unwrap() {
+        for c in syntax.chars().take(max_length).step_by(2) {
```
alamb, line 83: Previously the inner loop would iterate over all of the characters each time to find the `nth` one. 

I changed the loop so it just iterated over the characters directly. 

I believe this is responsible for most of the improvements in this PR

tpchgen/src/text.rs
```diff
@@ -80,14 +80,14 @@ impl TextPool {
...
+            match c {
                 'V' => Self::generate_verb_phrase(distributions, builder, random),
                 'N' => Self::generate_noun_phrase(distributions, builder, random),
                 'P' => {
                     let preposition = distributions.prepositions().random_value(random);
                     builder.append(preposition);
-                    builder.append(" the ");
+                    builder.append_bytes(b" the ");
```
alamb, line 90: Another thing I did that might help was to distinguish appending constant slices (where the length is known at compile time) and actual variable length strings to let the compiler generate specialized code when possible

scsmithr (approved):
lgtm

tpchgen/src/text.rs
clflushopt: My own accidentally quadratic whoopsie 🤦

clflushopt (approved):
LGTM

alamb:
Thanks @scsmithr and @clflushopt

PR #68, opened 2025-03-26
feat: remove chrono dependency

clflushopt:
- Closes #28 

In this change I replace the minimal usage of `chrono` by an inlined format routine for the purposes of building the date index, there seems to be minimal to no impact with regards to performance (as expected). `chrono::NaiveDate` has some nice defensive approaches to building from a `yyyy,mm,dd` tuple of `i32, i32, i32` in this case I didn't replicate any of the assertions or the checks; which given the fact that we know the space of dates we generate I assume is a safe trade off to make.

Once merged; we can confidently say that our little TPCH generator that could is officially dependency free !

Below are some timings I captured; feel free to reproduce.

```
clflushopt@workstation ~/P/tpchgen-rs (cl/feat/remove-chrono-dependency)> cargo build --release
   Compiling tpchgen v0.1.0 (/home/clflushopt/Projects/tpchgen-rs/tpchgen)
   Compiling tpchgen-cli v0.1.0 (/home/clflushopt/Projects/tpchgen-rs/tpchgen-cli)
    Finished `release` profile [optimized + debuginfo] target(s) in 3.64s
clflushopt@workstation ~/P/tpchgen-rs (cl/feat/remove-chrono-dependency)> time ./target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs/

________________________________________________________
Executed in    1.18 secs    fish           external
   usr time    4.34 secs    0.00 micros    4.34 secs
   sys time    0.56 secs  320.00 micros    0.56 secs

clflushopt@workstation ~/P/tpchgen-rs (cl/feat/remove-chrono-dependency)> git switch main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
clflushopt@workstation ~/P/tpchgen-rs (main)> cargo build --release
   Compiling num-traits v0.2.19
   Compiling iana-time-zone v0.1.62
   Compiling chrono v0.4.40
   Compiling tpchgen v0.1.0 (/home/clflushopt/Projects/tpchgen-rs/tpchgen)
   Compiling tpchgen-cli v0.1.0 (/home/clflushopt/Projects/tpchgen-rs/tpchgen-cli)
    Finished `release` profile [optimized + debuginfo] target(s) in 4.22s
clflushopt@workstation ~/P/tpchgen-rs (main)> time ./target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs/

________________________________________________________
Executed in    1.18 secs    fish           external
[...]

PR #69, opened 2025-03-26
Provide access to TPCHDates as (y,m,d)

alamb:
- Follow on to https://github.com/clflushopt/tpchgen-rs/pull/68
- Related to #46 

# Rationale
Basically I think having the translation from `i32` to (y,m,d) available for use is an important feature and I think I will use it when creating arrow arrays

I also took the opportunity to add some additional comments while I was making changes

I ran `time target/release/tpchgen-cli -s 1 --output-dir=/tmp/tpchdbgen-rs` and there is no change with this PR and on main


clflushopt (approved):
LGTM

alamb (commented):
I forgot to submit my comments -- sorry about that

PR #74, opened 2025-03-27
Implement arrow conversion for remaining tables

alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/46
- Follow on to https://github.com/clflushopt/tpchgen-rs/pull/71

1. Implements table --> Arrow conversion for the remaining TPCH tables
2. Adds test coverage

It is mostly a mechanical (but large) PR

### DIFF
--- a/tpchgen-arrow/src/conversions.rs
+++ b/tpchgen-arrow/src/conversions.rs
@@ -27,6 +27,7 @@ where
     let values = values.map(to_arrow_decimal);
     arrow::array::Decimal128Array::from_iter_values(values)
         .with_precision_and_scale(15, 2)
+        // safe to unwrap because 15,2 is within the valid range for Decimal128 (38)
         .unwrap()
 }
 
--- a/tpchgen-arrow/src/conversions.rs
+++ b/tpchgen-arrow/src/conversions.rs
@@ -1,5 +1,7 @@
 //! Routines to convert TPCH types to Arrow types
 
+use arrow::array::{StringViewArray, StringViewBuilder};
+use std::fmt::Write;
 use tpchgen::dates::TPCHDate;
 use tpchgen::decimal::TPCHDecimal;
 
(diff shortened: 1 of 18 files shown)


tpchgen/src/generators.rs
```diff
@@ -1186,16 +1186,16 @@ impl fmt::Display for PartSupp<'_> {
     }
 }
 
-/// Generator for PartSupplier table data
-#[derive(Debug)]
-pub struct PartSupplierGenerator<'a> {
+/// Generator for PartSupp table data
+#[derive(Debug, Clone)]
+pub struct PartSuppGenerator<'a> {
```
alamb, line 1191: I also renamed PartSupplier --> PartSupp to be consistent with TPCH terminology elsewhere

tpchgen-arrow/src/customer.rs
```diff
@@ -0,0 +1,121 @@
+use crate::conversions::{decimal128_array_from_iter, string_view_array_from_display_iter};
```
alamb, line 1: Basically I just applied the same pattern for all the remaining TPCH tables

tpchgen/src/generators.rs
```diff
@@ -1404,6 +1404,27 @@ impl fmt::Display for ClerkName {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd)]
```
alamb, line 1407: I also changed this from `char` to an enum so I could more easily convert it to StringViewArray 

I actually think I can apply the same idea for some of the status codes in LineItem and make it faster too. As a follow on PR

alamb:
@clflushopt  I think this is the last PR before I can add real parquet support

tpchgen-arrow/src/conversions.rs
```diff
@@ -1,5 +1,7 @@
 //! Routines to convert TPCH types to Arrow types
 
+use arrow::array::{StringViewArray, StringViewBuilder};
+use std::fmt::Write;
```
clflushopt, line 4: suggestion: this is mostly a personal preference so feel free to omit it but I would like it if we can keep imports sorted and grouped similar to how Go's default `fmt` does it (first block is `std` library, followed by local/internal crates and finally third party dependencies). Not important though mostly a personal taste/preference thing !

tpchgen-arrow/src/conversions.rs
```diff
@@ -17,6 +19,36 @@ pub fn to_arrow_date32(value: TPCHDate) -> i32 {
...
 
+/// Converts an iterator of TPCH decimals to an Arrow Decimal128Array
+pub fn decimal128_array_from_iter<I>(values: I) -> arrow::array::Decimal128Array
+where
+    I: Iterator<Item = TPCHDecimal>,
+{
+    let values = values.map(to_arrow_decimal);
+    arrow::array::Decimal128Array::from_iter_values(values)
+        .with_precision_and_scale(15, 2)
+        .unwrap()
```
clflushopt, line 30: question: can the `unwrap` fail in this case ? I don't see how since we control the values and we know "hand-wavey" that the input space fits in the decimal type, but I am asking because if the function is exposed do we need to communicate some kind of "failure modes" to the users ?

clflushopt (approved):
lgtm, my comments are non-blocking so feel free to merge ! this seems to have been cumbersone 😅 thanks for taking the initiative !!

tpchgen/src/generators.rs
alamb: Follow up the idea is here: https://github.com/clflushopt/tpchgen-rs/pull/75 (but still not complete)

tpchgen-arrow/src/conversions.rs
alamb: Make sense -- will fix. I think my editor (RustRover does  this and I don't really pay attention to the imports)

tpchgen-arrow/src/conversions.rs
alamb: In this case it can not fail -- the only way that `with_precision_and_scale` can fail is if the precison / scale are outside of the allowable range of arrow decimals (38 and 20 I believe). I will clarify this in comments

alamb (commented):
Thank you so much for the review @clflushopt

tpchgen-arrow/src/conversions.rs
alamb: done in 9a12ac03528e97b956d24c5be063fc0692f0dcd8

tpchgen-arrow/src/conversions.rs
alamb: in 9a12ac03528e97b956d24c5be063fc0692f0dcd8

alamb:
Here we go! On to parquet generation directly

PR #75, opened 2025-03-28
WIP optimize lineitem conversion to arrow

alamb:
WIP optimize lineitem conversion to arrow

This PR contains an attempt to skip the string representation and go directly to byte views 

TLDR is I found it didn't really make much difference. 

Another thing that might be more compelling would be to reuse allocations / buffers

But for now let's call it good enough

Main
```
$ time ./target/release/tpchgen-cli -s 10 --format=parquet --stdout > /dev/null

real	0m5.436s
user	0m55.109s
sys	0m8.402s
```

This PR
```
time ./target/release/tpchgen-cli -s 10 --format=parquet --stdout > /dev/null

real	0m5.461s
user	0m53.384s
sys	0m9.874s
```

### DIFF
--- a/tpchgen-arrow/src/conversions.rs
+++ b/tpchgen-arrow/src/conversions.rs
@@ -60,9 +60,8 @@ where
     let values = values.into_iter();
 
     // format all values directly into the buffer
-    let mut buffer: Vec<u8> = Vec::with_capacity (31*1024*1024);
+    let mut buffer: Vec<u8> = Vec::with_capacity(31 * 1024 * 1024);
     let mut views = Vec::with_capacity(values.size_hint().0);
-
 
     for v in values {
         use std::io::Write;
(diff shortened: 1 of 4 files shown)


clflushopt:
@alamb do we still want this as part of the next milestone or are we happy with current performance numbers ?

alamb:
> @alamb do we still want this as part of the next milestone or are we happy with current performance numbers ?

Well I would love better performance always, but this particular PR didn't seem to provide any measurable benefit so I don't think we should try and add it to the release

Another idea I had was to try making individual column iterators (would have to duplicate a significant part of the code from the row generators) -- and then make arrow directly from those column iterators. I don't plan to spend time working on that in the near term

clflushopt:
Going to close this for now, but I plan to do a big performance review soon and tinker about column oriented generation.

PR #83, opened 2025-03-29
feat: allow writing to stdout

clflushopt:
The goal of this change (Closes #77 ) is to allow writing to `stdout` which can be useful when we want to measure throughput via `pv` for example.

```
~/G/P/tpchgen-rs (cl/feat/write-to-stdout)> ./target/release/tpchgen-cli -s 100 --stdout | pv > /dev/null
 106GiB 0:01:07 [1.57GiB/s]
 ~/G/P/tpchgen-rs (cl/feat/write-to-stdout) [SIGINT|32]> ./target/release/tpchgen-cli -s 100 --stdout | pv -arb > /dev/null
 106GiB [1.36GiB/s] (1.36GiB/s)s)
```

Issue #77, opened 2025-03-28
[FEATURE] Feature to write to stdout  or /dev/null


alamb:
**Is your feature request related to a problem? Please describe.**
`tpchgen-cli` is often IO limited (it can generate data faster than disks can write) which makes it hard to gather repeatable benchmarks

@clflushopt suggested in a ticket I can't find right now it would be great to be able to generate tables )

**Describe the solution you'd like**
I would like some way to ignore the output

For example, a cli flag like
```shell
--dev-null
```

When specified would still create all the data but then not try and write it to a file, and would instead simply discard it

**Describe alternatives you've considered**

**Additional context**

PR #85, opened 2025-03-29
Support writing  parquet to stdout, document use of `pv`

alamb:
- Follow up to https://github.com/clflushopt/tpchgen-rs/pull/83
- Related to https://github.com/clflushopt/tpchgen-rs/issues/77

I would like to also calculate raw speed for parquet generation so it would be nice to support writing parquet to stdout as well

Also, @clflushopt  showed a great command to use `pv` to measure raw output speed which I wanted to add to the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,30 @@ Blazing fast TPCH benchmark data generator in pure Rust !
 
 (coming soon)
 
+## Measuring Performance
+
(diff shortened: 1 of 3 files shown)


clflushopt (approved):

alamb:
🚀

PR #86, opened 2025-03-29
fix: Use a different shorthand for parquet compression argument

clflushopt:
Addresses an issue where short arguments collide between `parts` and `parquet-compression`.


alamb (approved):
Thank you @clflushopt

Issue #81, opened 2025-03-29
[BUG] `"Parquet does not support more than 32767 row groups`  when generating SF1000


alamb:
**Describe the bug**
I am trying to create TPCH SF 1000 in a single parquet file but I get an error

**To Reproduce**

```
RUST_LOG=debug cargo run --release -- --format=parquet --scale-factor=1000 --tables=line-item
```

I get a panic like this:
```
thread 'tokio-runtime-worker' panicked at tpchgen-cli/src/parquet.rs:92:64:
called `Result::unwrap()` on an `Err` value: General("Parquet does not support more than 32767 row groups per file (currently: 32768)")
```

Entire example

```
alamb@aal-perf:/data/tpchgen-rs$ RUST_LOG=debug cargo run --release -- --format=parquet --scale-factor=1000 --tables=line-item
    Updating crates.io index
     Locking 179 packages to latest compatible versions
    Finished `release` profile [optimized + debuginfo] target(s) in 1.46s
     Running `target/release/tpchgen-cli --format=parquet --scale-factor=1000 --tables=line-item`
[2025-03-29T11:20:47Z DEBUG tpchgen_cli] Logging configured from environment variables
[2025-03-29T11:20:47Z DEBUG tpchgen_cli] Creating distributions and text pool
[2025-03-29T11:20:48Z INFO  tpchgen_cli] Created static distributions and text pools in 1.080727322s
[2025-03-29T11:20:48Z INFO  tpchgen_cli] Writing table lineitem (SF=1000) to lineitem.parquet
[2025-03-29T11:20:48Z DEBUG tpchgen_cli] Generating 32766 parts in total
[2025-03-29T11:20:48Z DEBUG tpchgen_cli::parquet] Generating Parquet with 22 threads, using SNAPPY compression

thread 'tokio-runtime-worker' panicked at tpchgen-cli/src/parquet.rs:92:64:
called `Result::unwrap()` on an `Err` value: General("Parquet does not support more than 32767 row groups per file (currently: 32768)")

```

**Expected behavior**
I expect the command to complete

alamb:
I am going to fix this

PR #87, opened 2025-03-29
Limit number of parquet row groups

alamb:
- Closes https://github.com/clflushopt/tpchgen-rs/issues/81

I verified I can generate parquet SF 1000 now without error using this command

I tested with this command:
```
RUST_LOG=debug cargo run --release -- -s 1000 --format=parquet --tables=line-item --stdout > /dev/null
```

With this fix, we will be in the position to actually benchmark and produce some charts :)

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -173,8 +173,6 @@ macro_rules! define_generate {
         async fn $FUN_NAME(&self) -> io::Result<()> {
             let filename = self.output_filename($TABLE);
             let (num_parts, parts) = self.parallel_target_part_count(&$TABLE);
-            // parquet files can have at most 32K row groups so cap the number of parts
-            let num_parts = num_parts.min(32767 - 1);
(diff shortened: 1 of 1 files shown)


tpchgen-cli/src/main.rs
```diff
@@ -383,7 +381,13 @@ impl Cli {
         };
         // target chunks of about 16MB (use 15MB to ensure we don't exceed the target size)
         let target_chunk_size_bytes = 15 * 1024 * 1024;
-        let num_parts = ((row_count * avg_row_size_bytes) / target_chunk_size_bytes) + 1;
+        let mut num_parts = ((row_count * avg_row_size_bytes) / target_chunk_size_bytes) + 1;
+
+        // parquet files can have at most 32767 row groups so cap the number of parts at that number
```
alamb, line 386: The limit needs to be applied here before generating the list of parts

clflushopt (approved):

Issue #76, opened 2025-03-28
[BUG] Growing Memory Use


alamb:
**Describe the bug**
the memory used by tpchgen-cli grows (slowly but steadily)

**To Reproduce**
- Needs #46 

```shell
tpchgen-cli -v --format=parquet --scale-factor=100  --output-dir=/tmp
```

And observe the memory usage -- it goes up for me

**Expected behavior**
I expect the memory to remain consitent during data generation

**Screenshots**

**Environment (please complete the following information):**
I saw this both on my Mac laptop and a debian linux machine

**Additional context**

I think the first thing to do would be to run a heap profile

clflushopt:
I will investigate and take care of this, thanks @alamb.

alamb:
I did some profiling this morning. I am pretty sure this only happens with parquet (not TBL):

## Memory Usage (TBL)
![Image](https://github.com/user-attachments/assets/b44be989-fe4d-48de-b003-7cdaa085fd25)


## Memory Usage (Parquet)
Here is the memory profile when writing parquet:
![Image](https://github.com/user-attachments/assets/b4a5ab8e-ec53-4809-9ad2-c5b54b3c497b)



BTW I was using XCode Instruments (and I had to grant access like https://github.com/rust-lang/rust/issues/107033)

clflushopt:
I started investigating this and captured a heap profile using `heaptrack` one thing to note is that the distributions hashmap allocated within the `LazyLock` **is not free-ed by Rust and instead by the OS**  as explained in https://docs.rs/rustc-std-workspace-std/latest/std/sync/struct.LazyLock.html#examples

![Image](https://github.com/user-attachments/assets/e3404143-2ca5-4ce9-a512-0b915f6e907b)

![Image](https://github.com/user-attachments/assets/3c690487-cd09-406b-818a-502f759f86c6)

Notice that the memory consumed in total is fairly stable, I am going to try and capture more profiles; the above was captured as :

```
clflushopt@workstation ~/P/tpchgen-rs (main)> heaptrack ./target/release/tpchgen-cli -v --format=parquet --scale-factor=100  --output-dir=/tmp
```

clflushopt:
I ran another heap profile this time using zSTD as a compression algorithm and did another pass with Parquet I am pretty sure the memory usage is due to the parquet writer.

![Image](https://github.com/user-attachments/assets/7ba6afa6-f5ce-42ba-ab6c-e93ce1c39b3c)

[...]

PR #88, opened 2025-03-29
Add documentation for zstd compression

alamb:
@clflushopt says in https://github.com/clflushopt/tpchgen-rs/issues/76#issuecomment-2764251375

> p.s: I think we should document how to pass ZSTD since it requires passing the level in a parenthesis format like this zstd(1) as such zstd\(1\) same goes for gzip which also requires a level.

So let's document it better!


clflushopt (approved):
lgtm

PR #92, opened 2025-03-30
fix: Make sure table name handling is done by the crate instead of clap

### LINKED ISSUE #91: [FEATURE] Allow specifying tables by their prefixes similar to `dbgen`
We want to be able to specify the list of tables to generate using their prefixes.

**Additional context**

clflushopt:
To allow supporting actual short hand we have to ensure that we use a custom TypedValueParser that is able to parse the refs correctly and align every abbrv with its long name and corresponding enum, one downside is that we lose some of the auto-derived nice things from `clap` like the list of options in the `--help` message due to a conflict from deriving `ValueEnum` vs implementing `FromStr` ourselves which is necessary to allow both abbrvs and long names.

I don't like this code because it's a lot of gymnastics for supporting command flags (I feel like clap is very much too powerful for our needs but maybe I am used to Go's stdlib provided argument handler).

There's also some non-trivial complexity in the implementation (side effect of how clap does things).

Closes #91

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -168,31 +168,6 @@ impl TypedValueParser for TableValueParser {
                 clap::builder::PossibleValue::new("PartSupp").help("PartSupp table (alias: S)"),
                 clap::builder::PossibleValue::new("Orders").help("Orders table (alias: O)"),
                 clap::builder::PossibleValue::new("LineItem").help("LineItem table (alias: L)"),
-                // Add aliases as separate possible values
-                clap::builder::PossibleValue::new("r")
-                    .help("Region table")
-                    .hide(true),
-                clap::builder::PossibleValue::new("n")
-                    .help("Nation table")
-                    .hide(true),
-                clap::builder::PossibleValue::new("s")
-                    .help("Supplier table")
-                    .hide(true),
-                clap::builder::PossibleValue::new("c")
-                    .help("Customer table")
-                    .hide(true),
-                clap::builder::PossibleValue::new("P")
-                    .help("Part table")
-                    .hide(true),
-                clap::builder::PossibleValue::new("S")
-                    .help("PartSupp table")
-                    .hide(true),
-                clap::builder::PossibleValue::new("O")
-                    .help("Orders table")
-                    .hide(true),
-                clap::builder::PossibleValue::new("L")
-                    .help("LineItem table")
-                    .hide(true),
             ]
             .into_iter(),
         ))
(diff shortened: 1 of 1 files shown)


tpchgen-cli/src/main.rs
```diff
@@ -133,6 +135,51 @@ impl Display for Table {
...
+        cmd: &clap::Command,
+        _: Option<&clap::Arg>,
+        value: &std::ffi::OsStr,
+    ) -> Result<Self::Value, clap::Error> {
+        let value = value
+            .to_str()
+            .ok_or_else(|| clap::Error::new(clap::error::ErrorKind::InvalidValue).with_cmd(cmd))?;
+        Table::from_str(value)
+            .map_err(|_| clap::Error::new(clap::error::ErrorKind::InvalidValue).with_cmd(cmd))
+    }
```
alamb, line 156: If looks like we could implement `possible_values` to get a list of possible values too: 

https://docs.rs/clap/latest/clap/builder/trait.TypedValueParser.html#method.possible_values

```rust
    fn possible_values(&self) -> Option<Box<dyn Iterator<Item=PossibleValue> + '_>> {
        todo!()
    }
```

tpchgen-cli/src/main.rs
```diff
@@ -133,6 +135,51 @@ impl Display for Table {
...
+}
+
+impl FromStr for Table {
+    type Err = &'static str;
+
+    /// Returns the table enum value from the given string full name or abbreviation
+    ///
+    /// The original dbgen tool allows some abbreviations to mean two different tables
+    /// like 'p' which aliases to both 'part' and 'partsupp'. This implementation does
+    /// not support this since it just adds unnecessary complexity and confusion so we
```
alamb, line 166: 💯

alamb (approved):
Thanks @clflushopt !

> I don't like this code because it's a lot of gymnastics for supporting command flags (I feel like clap is very much too powerful for our needs but maybe I am used to Go's stdlib provided argument handler).
> 
> There's also some non-trivial complexity in the implementation (side effect of how clap does things).

Thanks @clflushopt  -- I didn't quite follow if you thought this PR was to complicated or if the clap implementation was too complicated. Is there anything you think we should be changing anything about this complexity?

tpchgen-cli/src/main.rs
clflushopt: fixed !

clflushopt:
@alamb The PR itself is not complicated, my comment was specifically about the ergonomics of `clap` but I guess it's powerful for a reason (less stuff for us to implement !).

PR #94, opened 2025-03-30
fix: pin local crate versions

clflushopt:
To make sure version compatibility stays within what we expect I am pinning the `tpchgen`crate versions for `tpchgen-cli` and `tpchgen-arrow`


alamb:
Thanks @clflushopt

PR #95, opened 2025-03-30
Add `as_str` helper to `OrderStatus`

scsmithr:
Adds a helper method for returning a static string ("F", "O", "P") for `OrderStatus` to allow getting the string representation without needing to format it.


clflushopt (approved):
lgtm

Issue #60, opened 2025-03-22
Tracking issue for v0.1.0


clflushopt:
This is a tracking issue for what I expect will be a v1.0.0 release that we can publish and make available as an extension in [Datafusion](https://github.com/apache/datafusion) and eventually if it gets accepted as part of [Datafusion Community Extensions](https://github.com/datafusion-contrib).

Since I won't be available to review, finish up some PRs (#5 and #28) and help with others until March 24th I will keep this under "nice things to have" for now.

* Pull requests I would like to have in the first milestone
  * #56 
  * #53 
  * #46 
  * #28 
  * #10 
  * #5 
  * https://github.com/clflushopt/tpchgen-rs/issues/65

Otherwise; I won't file issues for them but the following still need to be addressed :

* Improve project and code documentation beyond what we have planned in #24.
* Take a look and eventually contribute to the draft for #45 
* Coordinate with @alamb to have a PoC for an extension in Datafusion
* Run conformance tests against Java & C implementation for scale factors > 1
* Compare performance against DuckDB for scale factors greater than 10 (see https://duckdb.org/docs/stable/extensions/tpch.html#resource-usage-of-the-data-generator)
* Ensure crate-ability of the existing package to push it to `crates.io`.

clflushopt:
@scsmithr @alamb let me know if you have any wishlist items you want included for the first release

alamb:
This looks like a good (though ambitious list).

My opinion is that  this one is nice to have but not required (given how prevalent chrono is)
- https://github.com/clflushopt/tpchgen-rs/issues/28


I plan to work on the following
- https://github.com/clflushopt/tpchgen-rs/issues/53
- https://github.com/clflushopt/tpchgen-rs/issues/46
- https://github.com/clflushopt/tpchgen-rs/issues/45
- https://github.com/clflushopt/tpchgen-rs/issues/24
- https://github.com/clflushopt/tpchgen-rs/issues/10

niebayes:
@alamb I just confirmed an issue does exist for reading csv files through DataFusion. Please check it: https://github.com/clflushopt/tpchgen-rs/pull/66#issuecomment-2754881275

alamb:
> [@alamb](https://github.com/alamb) I just confirmed an issue does exist for reading csv files through DataFusion. Please check it: [#66 (comment)](https://github.com/clflushopt/tpchgen-rs/pull/66#issuecomment-2754881275)

[...]

PR #97, opened 2025-03-30
feat: add to_unix_epoch helper to TPCHDate

clflushopt:
Addresses a feature request by @scsmithr in #60 see [comment](https://github.com/clflushopt/tpchgen-rs/issues/60#issuecomment-2764785631).

The date conversion logic is largely mechanical **but it will compute everything relative to midnight**.

### DIFF
--- a/tpchgen/src/dates.rs
+++ b/tpchgen/src/dates.rs
@@ -97,8 +97,14 @@ impl Display for TPCHDate {
 }
 
 impl TPCHDate {
-    /// Number of days between TPC-H epoch and the Unix epoch (1970-01-01).
-    const UNIX_EPOCH_OFFSET: i32 = 8035;
+    /// Number of days that must be added to a TPCH date to get a Unix epoch
+    /// relative date.
+    ///
+    /// * Arrow `Date32` are days since the epoch (1970-01-01)
+    /// * [`TPCHDate`]s are days since MIN_GENERATE_DATE (1992-01-01)
+    ///
+    /// This value is `8035` because `1992-01-01` is `8035` days after `1970-01-01`
+    pub const UNIX_EPOCH_OFFSET: i32 = 8035;
 
     /// Create a new TPCHDate from a generated date
     pub fn new(generated_date: i32) -> Self {
--- a/tpchgen/src/dates.rs
+++ b/tpchgen/src/dates.rs
@@ -131,7 +131,7 @@ impl TPCHDate {
     }
 
     /// Returns unix epoch time for this date.
-    fn to_unix_epoch(&self) -> i64 {
+    pub fn to_unix_epoch(&self) -> i64 {
         let (y, m, dy) = self.to_ymd();
         days_since_unix_epoch(1900 + y, m, dy) * 86400
     }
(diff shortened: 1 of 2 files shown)


tpchgen/src/dates.rs
```diff
@@ -129,6 +129,12 @@ impl TPCHDate {
     fn to_julian(date: i32) -> i32 {
         JULIAN_DATE[(date - MIN_GENERATE_DATE) as usize]
     }
+
+    /// Returns unix epoch time for this date.
```
alamb, line 133:
```suggestion
    /// Returns unix epoch time: number of seconds since `1970-01-01`
```

tpchgen/src/dates.rs
```diff
@@ -222,4 +255,23 @@ mod test {
             assert_eq!(format_ymd(y, m, dy), date.to_string());
         }
     }
+
```
alamb, line 258: I did a quick double check and it looks good to me 👍 

Screenshot 2025-03-31 at 6 34 07 AM

alamb (approved):
Thanks @clflushopt

scsmithr (commented):
Tbh I don't know how useful returning the unix time in seconds is. I was more thinking that this would return an i32 representing days since unix epoch (e.g. something like what the arrow conversion is doing here: https://github.com/clflushopt/tpchgen-rs/blob/main/tpchgen-arrow/src/conversions.rs#L16-L20).

Also I'd want to avoid having the compute the days since epoch every time since running this on thousands of rows could add up -- the arrow conversion helper just adds a relative constant which I think is the way to go.

clflushopt:
> Tbh I don't know how useful returning the unix time in seconds is. I was more thinking that this would return an i32 representing days since unix epoch (e.g. something like what the arrow conversion is doing here: https://github.com/clflushopt/tpchgen-rs/blob/main/tpchgen-arrow/src/conversions.rs#L16-L20).
> 
> Also I'd want to avoid having the compute the days since epoch every time since running this on thousands of rows could add up -- the arrow conversion helper just adds a relative constant which I think is the way to go.

Alright, I think I misunderstood the original ask, I assumed it was for relative Unix epochs (which are usually seconds) not days. That's easier I can make the quick fix.

tpchgen/src/dates.rs
```diff
@@ -97,6 +97,9 @@ impl Display for TPCHDate {
 }
 
 impl TPCHDate {
+    /// Number of days between TPC-H epoch and the Unix epoch (1970-01-01).
```
alamb, line 100: As this now duplicates the `TPCHDATE_TO_DATE32_OFFSET` and `to_arrow_date32`   https://github.com/clflushopt/tpchgen-rs/blob/c4d12fa0be8cbc2e4d2a221204c8f5ed1368d484/tpchgen-arrow/src/conversions.rs#L53-L69

Perhaps you could move the code / comments / tests from tpchgen-arrow to this module and remove the copy there, and call it directly?

alamb (approved):
Thanks @clflushopt  -- I think this looks good to me

I think it would be cool to avoid the replication with tpchgen-arrow but also it is not a huge deal either

tpchgen/src/dates.rs
clflushopt: Sounds good !

PR #102, opened 2025-04-02
chore: tweak doc comments, add `inline` for good measure

alamb:
- Follow on to https://github.com/clflushopt/tpchgen-rs/pull/97

The comments still refer to to Arrow even though the code is now in tpchgen, so let's tweak the comments

### DIFF
--- a/tpchgen/src/dates.rs
+++ b/tpchgen/src/dates.rs
@@ -141,6 +141,7 @@ impl TPCHDate {
 
     /// Returns the number of days since the Unix epoch this date
     /// represents.
+    #[inline(always)]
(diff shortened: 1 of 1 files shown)


tpchgen/src/dates.rs
```diff
@@ -141,6 +141,7 @@ impl TPCHDate {
 
     /// Returns the number of days since the Unix epoch this date
     /// represents.
+    #[inline(always)]
```
alamb, line 144: I am not sure this helps but I don't think it hurts either

clflushopt (approved):

clflushopt:
Thanks for catching this one @alamb

PR #106, opened 2025-04-05
fix: bump tpchgen-arrow and tpchgen-cli to 0.1.1

clflushopt:


alamb (approved):
Thanks @clflushopt 

I suggest we maybe even want to change the version to `1.0`

PR #109, opened 2025-04-07
Improve documentation on additional shasum checks

alamb:
I have nightmares about getting incorrect results so have been doing some additional verification and gathered some more data. I figured I would post it here


kevinjqliu (approved):
LGTM

Do you think its worth adding a CI check comparing with the checksum for SF1?

alamb:
> LGTM
> 
> Do you think its worth adding a CI check comparing with the checksum for SF1?

Thanks @kevinjqliu  -- I think the existing integration test actually already do this (they actually compare using `diff`) for example

https://github.com/clflushopt/tpchgen-rs/blob/cacff017fd8c88f76b6a9df8b35b7049c078c4cb/tests/conformance.sh#L44

kevinjqliu:
awesome, i see this is already part of the [Github workflow](https://github.com/clflushopt/tpchgen-rs/blob/cacff017fd8c88f76b6a9df8b35b7049c078c4cb/.github/workflows/rust.yml#L75) and in fact ran for this PR https://github.com/clflushopt/tpchgen-rs/actions/runs/14310557458/job/40104376355?pr=109

PR #110, opened 2025-04-08
Correctly pin chrono

matthewmturner:


tpchgen-arrow/Cargo.toml
```diff
@@ -9,9 +9,9 @@ readme = "README.md"
...
 [dependencies]
-arrow = { version = "54.3.0", default-features = false, features = ["prettyprint"] }
+arrow = { version = "54.1.0", default-features = false, features = ["prettyprint"] }
 tpchgen = { path = "../tpchgen", version = "0.1.1" }
 
 [dev-dependencies]
-arrow-csv = "54.3.0"
-chrono = "0.4.39"
+arrow-csv = "54.1.0"
+chrono = "=0.4.39"
```
matthewmturner, line 17: Need the `=` to actually pin it otherwise its implicitly using a `^`

tpchgen-arrow/Cargo.toml
```diff
@@ -9,9 +9,9 @@ readme = "README.md"
 license = "Apache-2.0"
 
 [dependencies]
-arrow = { version = "54.3.0", default-features = false, features = ["prettyprint"] }
+arrow = { version = "54.1.0", default-features = false, features = ["prettyprint"] }
```
alamb, line 12: Maybe we can make it even more compatible with '54.0.0'

tpchgen-arrow/Cargo.toml
alamb: I think it would be better if downstream crates / users pinned chrono (otherwise using this crate will force a particular version of chrono which doesn't seem quite right

PR #127, opened 2025-04-17
feat: fix CLI option docs

kevinjqliu:
Fix up option docs based on Parser:
* `--scale` -> `--scale-factor`
* `-O, --output` -> `-o, --output-dir`
* `-F` -> `-f`

Also fixed version info:
```
tpchgen-cli -V
tpchgen-cli --version
```

Test
```
cargo run --bin tpchgen-cli -- -V
cargo run --bin tpchgen-cli -- --version
```

PR #128, opened 2025-04-18
feat: fix CI and run CI for every PR

kevinjqliu:
Format issue in #127 was not caught in CI but was caught when merging. 

It looks like the github action was not run on PR. Let's fix that too

According to claude, the github action was not triggered because it [specifies `main` branch](https://github.com/clflushopt/tpchgen-rs/blob/928d57af42b3752bd7f8c211f603337611d0241f/.github/workflows/rust.yml#L6-L7) but the PR wants to merge into ` clflushopt:main`
```
kevinjqliu wants to merge 2 commits into clflushopt:main from kevinjqliu:kevinjqliu/fix-ci  
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: ["main"]
   pull_request:
-    branches: ["main"]
   workflow_dispatch:  # Allow manual triggering
 
 env:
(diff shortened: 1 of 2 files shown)


clflushopt (approved):
Thanks for catching this one, lgtm !

PR #138, opened 2025-04-21
Replace ByteArrayBuilder with Vec

scsmithr:
Replaces `ByteArrayBuilder` with a plain old `Vec`.

The goal was faster text pool creation. I thought avoiding zeroing out the vec would be more significant. But there's almost no change. I do think using a `Vec` and reducing some of the abstraction around this can help future optimization work though.

Vast majority of the time is still spent on `memmove`. I think there's a path towards doing fewer larger batches for `memmove` instead of the many small calls we're doing now. But that'd need to be in a followup at some point.

Perf changes within noise threshold. Ran three times with each branch with small sf to test text pool initialization.

```
time target/release/tpchgen-cli -s 0.01 --output-dir=/tmp/tpchdbgen-rs
```

| main | this branch |
|-|-|
| 1.017 secs, 1.022 secs, 1.023 secs | 1.008 secs, 1.019 secs, 1.013 secs|

### DIFF
--- a/tpchgen/src/distribution.rs
+++ b/tpchgen/src/distribution.rs
@@ -14,7 +14,12 @@ pub struct Distribution {
     name: &'static str,
     values: Vec<&'static str>,
     weights: Vec<i32>,
-    distribution: Option<Vec<&'static str>>,
+    /// Weighted text values generated from the distribution.
+    ///
+    /// If the table this distribution is for isn't actually a valid
+    /// distribution (e.g. 'nation'), then this vec will be empty, and
+    /// `max_weight` set to -1.
+    distribution: Vec<&'static str>,
     max_weight: i32,
 }
 
@@ -49,9 +54,9 @@ impl Distribution {
                 }
             }
 
-            (Some(dist), max)
+            (dist, max)
         } else {
-            (None, -1)
+            (Vec::new(), -1)
         };
 
         let values = distribution.into_iter().map(|(tok, _)| tok).collect();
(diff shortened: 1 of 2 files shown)


tpchgen/src/distribution.rs
```diff
@@ -14,7 +14,12 @@ pub struct Distribution {
     name: &'static str,
     values: Vec<&'static str>,
     weights: Vec<i32>,
-    distribution: Option<Vec<&'static str>>,
+    /// Weighted text values generated from the distribution.
+    ///
+    /// If the table this distribution is for isn't actually a valid
+    /// distribution (e.g. 'nation'), then this vec will be empty, and
+    /// `max_weight` set to -1.
+    distribution: Vec<&'static str>,
```
scsmithr, line 22: I removed the `Option` here since it was only telling us if this distribution is a valid distribution or not, but we can already glean that information from the `max_weight` being -1.

This removes the `Option` check in `random_value`.

tpchgen/src/text.rs
```diff
@@ -41,25 +41,22 @@ impl TextPool {
 
     /// Returns a new text pool with a predefined size and set of distributions.
     pub fn new(size: i32, distributions: &Distributions) -> Self {
-        let mut output = ByteArrayBuilder::new(size as usize + Self::MAX_SENTENCE_LENGTH as usize);
         let mut rng = RowRandomInt::new(933588178, i32::MAX);
+        let mut text_bytes = Vec::with_capacity(size as usize + Self::MAX_SENTENCE_LENGTH as usize);
```
scsmithr, line 45: `with_capacity` avoids (explicitly) zeroing out the memory before writing to it.

I didn't see much of difference on mac, but I think mac will zero out memory anyways. I'm curious to see if this would speed things up on linux.

scsmithr:
cc @alamb @clflushopt

clflushopt (approved):
LGTM good stuff ! Thanks @scsmithr

PR #153, opened 2025-07-03
bump arrow version to match datafusion

joseph-isaacs:


clflushopt:
Thanks for the pull request @joseph-isaacs and to answer @alamb yes I plan to merge the row group fix and the parallel single part trick in the coming week under a new tag.

Issue #80, opened 2025-03-29
[FEATURE] Mutli-threaded generation of specific parts (when `--part` and `--parts` are specified)


alamb:
**Is your feature request related to a problem? Please describe.**
When `--part` and `--parts` are specified only a single core is used to generate the results.

The TPCH CLI command line allows generating only parts of the dataset

For example this command only uses one core to generate 1/10th of the data for line item (part 1 of 10)

```shell
tpchgen-cli --scale-factor=100  --tables=line-item --part 1 --parts 10
```

However, this command to make the entire table actually completes faster than creating the single part (because it is multi-threaded)

```shell
tpchgen-cli --scale-factor=100  --tables=line-item 
```

**Describe the solution you'd like**

I would like the part generation to be multithreaded as well

**Describe alternatives you've considered**

I think this function needs to be adjusted: https://github.com/clflushopt/tpchgen-rs/blob/39b125f2e09ebd5ba0b55df243170d6ba3947b81/tpchgen-cli/src/main.rs#L334-L333

**Additional context**

alamb:
FYI @niebayes / @clflushopt  this might be an interesting task as well

djouallah:
2h 30 minutes to generate 1 TB using 2 cores and 16 GB of RAM, that's not bad at all, still multi threading will be nice though :)

![Image](https://github.com/user-attachments/assets/0617bdfc-07ca-4de5-b165-45bbd1fcd5bb)

PR #151, opened 2025-06-30
Alamb/smaller row groups

alamb:
Related to https://github.com/clflushopt/tpchgen-rs/issues/150

This is a PR to test if making smaller row groups solves the problem reported in https://github.com/clflushopt/tpchgen-rs/issues/150

PR #155, opened 2025-07-14
feat: Chunkify single parts to generate them in parallel

clflushopt:
- Closes https://github.com/clflushopt/tpchgen-rs/issues/80

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -417,9 +417,20 @@ impl Cli {
     /// - num_parts is the total number of parts to generate
     /// - part_list is the list of parts to generate (1 based)
     fn parallel_target_part_count(&self, table: &Table) -> (i32, Vec<i32>) {
-        // parallel generation disabled if user specifies a part explicitly
+        // If a single part is specified, split it into chunks to enable parallel generation.
         if self.part != 1 || self.parts != 1 {
-            return (self.parts, vec![self.part]);
+            // Heuristic: 4 chunks per thread to maximize throughput.
+            let num_chunks = (self.num_threads * 4) as i32;
+
+            // The new total number of parts is the original number of parts multiplied by the number of chunks.
+            let new_total_parts = self.parts * num_chunks;
+
+            // The new part numbers to generate are the chunks that make up the original part.
+            let start_part = (self.part - 1) * num_chunks + 1;
+            let end_part = self.part * num_chunks;
+            let new_parts_to_generate = (start_part..=end_part).collect();
+
+            return (new_total_parts, new_parts_to_generate);
         }
 
         // Note use part=1, part_count=1 to calculate the total row count
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -91,11 +91,11 @@ struct Cli {
     tables: Option<Vec<Table>>,
 
     /// Number of parts to generate (manual parallel generation)
-    #[arg(short, long, default_value_t = 1)]
+    #[arg(short, long, default_value_t = -1)]
     parts: i32,
 
     /// Which part to generate (1-based, only relevant if parts > 1)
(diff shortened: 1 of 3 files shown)


tpchgen-cli/src/main.rs
```diff
@@ -417,9 +417,20 @@ impl Cli {
     /// - num_parts is the total number of parts to generate
     /// - part_list is the list of parts to generate (1 based)
     fn parallel_target_part_count(&self, table: &Table) -> (i32, Vec<i32>) {
-        // parallel generation disabled if user specifies a part explicitly
+        // If a single part is specified, split it into chunks to enable parallel generation.
         if self.part != 1 || self.parts != 1 {
-            return (self.parts, vec![self.part]);
+            // Heuristic: 4 chunks per thread to maximize throughput.
```
alamb, line 422: I don't think there is any reason to use more tasks than threads. In other words, I recommend using `num_chunks = self.num_threads`)

I tested locally and using `self.num_threads` kept all the cores busy.

alamb (approved):
Looks good to me

I ran with
```shell
cargo run --release -- --tables lineitem --scale-factor 1000 --part 1 --parts 100 --format=parquet
```

And it kept all my cores busy as promised. 

Thank yoU @clflushopt

alamb:
I made a PR to add some integration tests here
- https://github.com/clflushopt/tpchgen-rs/pull/151

I think https://github.com/clflushopt/tpchgen-rs/pull/156  covers the correctness  aspect of this PR

I am not sure it covers the "make it parallel" functional aspect, but then again I am not sure that is possible

tpchgen-cli/src/main.rs
alamb: Thank you @clflushopt 

I think the integration tests are the main thing that is missing in general

It is also the thing that is missing from
- https://github.com/clflushopt/tpchgen-rs/pull/151

alamb: I ran these integration tests and sadly they fail with this change (as you reporeted)

alamb: FWIW I didn't carefully investigate what is different between the expected output and actual output file (e.g. is it an off by one error or something). I think it is worth exploring that -- I will do so

tpchgen-cli/src/plan.rs
```diff
@@ -66,12 +67,49 @@ impl GenerationPlan {
...
         cli_part: i32,
         cli_part_count: i32,
+        num_threads: usize,
     ) -> Self {
-        // parallel generation disabled if user specifies a part explicitly
+        // If a single part is specified, split it into chunks to enable parallel generation.
         if cli_part != -1 || cli_part_count != -1 {
+            // These tables are small not parameterized by part count,
+            // so we must create only a single part.
+            if table == &Table::Nation || table == &Table::Region {
```
alamb, line 76: this was the bug in previous versions -- Nation and Region can't be made with parts and so if we do that it gets

tpchgen-cli/src/main.rs
```diff
@@ -91,11 +91,11 @@ struct Cli {
     tables: Option<Vec<Table>>,
 
     /// Number of parts to generate (manual parallel generation)
-    #[arg(short, long, default_value_t = 1)]
+    #[arg(short, long, default_value_t = -1)]
```
alamb, line 94: If I didn't make this change, then tpchgen-cli was always generating at most 16 parts...

tpchgen-cli/src/plan.rs
clflushopt: Alright after a quick review I realized what had happened with the duplicate data.

clflushopt:
Sounds good @alamb did a sanity check at scale factor 10 with master it all looks good !

clflushopt (commented):
Can be merged once you approve @alamb thanks for picking this one up !

alamb (approved):
Thanks @clflushopt

PR #156, opened 2025-07-15
feat: Add integration tests for `tpchgen-cli`

alamb:
- Related to  https://github.com/clflushopt/tpchgen-rs/pull/155
- Related to https://github.com/clflushopt/tpchgen-rs/pull/151

# Rationale
As we make changes to tpchgen-cli, we need some way to test that our changes are correct.
Let's add some tests

# Changes
Add CLI based tests using `assert_cmd` 

THis was partly written by GPT 4.0 / Github copilot


mrendi29:
Thanks @alamb  for taking a stab at this. @kevinjqliu  i recall you had some test that you had not pushed yet, so question to you both, do we want to build off the integration tests suite within this branch/PR or should we split this into multiple PRs?

clflushopt (approved):
Thanks @alamb I am approving this PR to unblock and will let you resolve the comments from @mrendi29. As a follow up, do we want to have `/data/` use two sources of truth `duckdb + parquet` and `dbgen` outputs ?

alamb:
> Thanks @alamb I am approving this PR to unblock and will let you resolve the comments from @mrendi29. As a follow up, do we want to have `/data/` use two sources of truth `duckdb + parquet` and `dbgen` outputs ?

Thank you @clflushopt  🙏 

When you say "two sources of truth" I wonder what you mean? Do you mean some way to verify that the actual values written into the .parquet are are correct?

The tests in this PR do not do that (as you note)

Issue #154, opened 2025-07-07
[DISCUSS] v2.0.0


kevinjqliu:
### Open issues related to v2.0.0
- [x] Memory size growth (#76 & #150)
- [x] #152
- [x] #146
- [x] #80
- [ ] (not sure if we want to include this one) #145 

We dont have to resolve all of these, I think its good to see everything in one page :) 
Lets figure out which ones to include 

### Unrelated to v2.0.0
- #103
- #51

kevinjqliu:
One of my wishlist item is to be able to configure 
- row group size
- file size 

We're already working on making row group size configurable 
And I think writing configuring target file size (i.e. writing multiple files) might be better suited to implement in https://github.com/clflushopt/datafusion-tpch

clflushopt:
I will have a pull request for parallelizing specific parts with tests soonish and I can potentially address both #103 and #145 as well.

alamb:
> I will have a pull request for parallelizing specific parts with tests soonish and I can potentially address both [#103](https://github.com/clflushopt/tpchgen-rs/issues/103) and [#145](https://github.com/clflushopt/tpchgen-rs/issues/145) as well.

Update; the PR is here: https://github.com/clflushopt/tpchgen-rs/pull/155

alamb:
Here are some tests to help with finishing up these features
- https://github.com/clflushopt/tpchgen-rs/pull/156

alamb:
I spent some time reviewing the various issues like
- https://github.com/clflushopt/tpchgen-rs/issues/80
- https://github.com/clflushopt/tpchgen-rs/issues/152
- https://github.com/clflushopt/tpchgen-rs/issues/146

And I think they are all somewhat related

## Background
At a high level, parallelization in `tpchgen-cli` is based on "parts". A "part" is a logical chunk of the overall output, and the data generators can create parts individually. 

For `tbl` and `csv` generation, the system generates `num-threads` parts in parallel, flusing them to output 

When writing parquet files, the output file has one row group for each "part"

Thus I think we can solve / fix pretty much all the issues above by changing the algorithm for deciding parts, which is here

https://github.com/clflushopt/tpchgen-rs/blob/4322f584610ecde88977cd5dc9f08057bf920c87/tpchgen-cli/src/main.rs#L420-L419

## Constraints on parts

[...]

Issue #152, opened 2025-07-01
[BUG] Partitioned writes produce parquet files with a single row group.


mrendi29:
**Describe the bug**
When using partitioned writes, the parquet writer produces files that have a single row group.
We initially thought this was causing #150 , but after @alamb had an initial working version on #151 we noticed that it did not affect the original issue. 

@alamb has already a patch on #151 , it will be great if we can merge that. 

**To Reproduce**
Run tpchgen-cli with a partitioned write. You will see the parquet file has 1 row group: 
```
root@localhost:/mnt/test-tpch/tpch-sf750000# parquet-tools inspect lineitem.parquet | head

############ file meta data ############
created_by: parquet-rs version 54.3.1
num_columns: 16
num_rows: 6001215
num_row_groups: 1
format_version: 1.0
serialized_size: 2269
```

**Expected behavior**
Parquet writer should produce files with multiple row groups/

alamb:
I think we are in a better position here on main , but still not great -- main will create a parquet file with a number of row groups that is the same as the number of cores, which is not ideal either

I have an idea of how to fix this

alamb:
I am quite pleased with the result / PR here:
- https://github.com/clflushopt/tpchgen-rs/pull/168

Issue #146, opened 2025-05-21
[FEATURE] add an option for row group size


djouallah:
the current row group is around 187K rows, which is rather small specially for bigger SF, it will be nice to have an option to customize it.

james-willis:
Should the row group size be set by # of rows or size in bytes?

In my experience, generally row group size is described in mb. For example the parquet docs: https://parquet.apache.org/docs/file-format/configurations/


Spark and iceberg I believe also target a size in mb for their parquet writers.

clflushopt:
We can target the size in bytes since we know approximately the size of rows in the TPCH dataset.

alamb:
> In my experience, generally row group size is described in mb. For example the parquet docs: https://parquet.apache.org/docs/file-format/configurations/
> 
> Spark and iceberg I believe also target a size in mb for their parquet writers.

BTW the flag we merged in 
- https://github.com/clflushopt/tpchgen-rs/pull/158

Is in terms of MB

PR #157, opened 2025-07-15
refactor: Extract plan generation to `GenerationPlan`, add docs and tests

alamb:
- related to https://github.com/clflushopt/tpchgen-rs/issues/80
- related to https://github.com/clflushopt/tpchgen-rs/issues/152
- related to https://github.com/clflushopt/tpchgen-rs/issues/146

# Rationale
As described on https://github.com/clflushopt/tpchgen-rs/issues/154#issuecomment-3073677672 the generation of TPCH Gen "parts" defines how the resulting files are laid out as well as the parallelism

However, the current code is not really tested. As we begin to start modifying this logic we need some way to test it.

# Changes
1. Extract the calculation of parts into its own struct (to make testing feasible)
2. Add docs that explains more about what is going on 
3. Add unit test coverage of part calculation

This will certainly conflict with https://github.com/clflushopt/tpchgen-rs/pull/155 from @clflushopt

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -42,12 +42,14 @@
 mod csv;
 mod generate;
 mod parquet;
+mod plan;
 mod statistics;
 mod tbl;
 
 use crate::csv::*;
 use crate::generate::{generate_in_chunks, Sink, Source};
 use crate::parquet::*;
+use crate::plan::GenerationPlan;
 use crate::statistics::WriteStatistics;
 use crate::tbl::*;
 use ::parquet::basic::Compression;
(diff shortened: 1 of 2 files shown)


tpchgen-cli/src/main.rs
```diff
@@ -411,69 +419,6 @@ impl Cli {
         File::create(path)
     }
 
-    /// Returns a list of "parts" (data generator chunks, not TPCH parts) to create
```
alamb, line 414: I simply moved this logic into `GenerationPlan::new()` -- I tried to keep the algorithm the same in this PR (though I expect we will change the logic in subsequent PRs)

tpchgen-cli/src/plan.rs
```diff
@@ -0,0 +1,351 @@
...
+    /// to specify a particular part or number of parts to generate.
+    pub fn new(
+        table: &Table,
+        format: OutputFormat,
+        scale_factor: f64,
+        cli_part: i32,
+        cli_part_count: i32,
+    ) -> Self {
+        // parallel generation disabled if user specifies a part explicitly
+        if cli_part != -1 || cli_part_count != -1 {
```
alamb, line 71: as mentioned above, this logic is (meant) to be unchanged

tpchgen-cli/src/plan.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+}
+
+impl Display for GenerationPlan {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "GenerationPlan for {} parts", self.part_count)
+    }
+}
+
+#[cfg(test)]
```
alamb, line 149: Here is the real reason for the change -- to be able to write these tests

alamb:
Thank you @clflushopt  

FYI @kevinjqliu  -- this might make the row group resizing code eaiser (now it becomes a function of modifing the generation plan)

tpchgen-cli/src/plan.rs
```diff
@@ -0,0 +1,349 @@
...
+
+/// Converts the `GenerationPlan` into an iterator of (num_parts, part_number)
+impl IntoIterator for GenerationPlan {
+    type Item = (i32, i32);
+    type IntoIter = std::vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.part_list
+            .into_iter()
+            .map(|part_number| (self.part_count, part_number))
```
alamb, line 137: For anyone following along this is a bug, it should be

```suggestion
            .map(|part_number| (part_number, self.part_count))
```

I fixed it in
https://github.com/clflushopt/tpchgen-rs/pull/156

PR #158, opened 2025-07-15
feat: make parquet row groups size configurable 

kevinjqliu:
- closes https://github.com/clflushopt/tpchgen-rs/issues/146

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -95,7 +95,7 @@ struct Cli {
     #[arg(short, long)]
     parts: Option<i32>,
 
-    /// Which part(tion) to generate (1-based)
+    /// Which part(ition) to generate (1-based)
     ///
     /// If not specified, generates all parts
     #[arg(long)]
@@ -133,14 +133,14 @@ struct Cli {
     #[arg(long, default_value_t = false)]
     stdout: bool,
 
-    /// Target size in bytes row group in Parquet files
+    /// Target size in row group bytes in Parquet files
     ///
     /// Row groups are the typical unit of parallel processing and compression
     /// in Parquet. With many query engines, smaller row groups enable better
     /// parallelism and lower peak memory use but may reduce compression
     /// efficiency.
     ///
-    /// Note parquet files are limited to 32k row groups, so at high scale
+    /// Note: parquet files are limited to 32k row groups, so at high scale
     /// factors, the row group size may be increased to keep the number of row
     /// groups under this limit.
     ///


tpchgen-cli/src/main.rs
```diff
@@ -19,6 +19,7 @@
 //!         --part <N>               Which part to generate (1-based, default: 1)
 //!     -n, --num-threads <N>        Number of threads to use (default: number of CPUs)
 //!     -c, --parquet-compression <C> Parquet compression codec, e.g., SNAPPY, ZSTD(1), UNCOMPRESSED (default: SNAPPY)
+//!         --parquet-row-group-size <N> Number of rows per row group in Parquet files (default: 1048576)
```
alamb, line 22: this looks great

tpchgen-cli/src/parquet.rs
```diff
@@ -16,6 +16,9 @@ use std::sync::Arc;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tpchgen_arrow::RecordBatchIterator;
 
+/// Type alias for a collection of row groups, where each row group contains [`ArrowColumnChunk`]s
```
alamb, line 19: Thank you @kevinjqliu

I am worried that this code doesn't account for the 32k limit on the number of row groups nor how the parts work

I tested in two configuration:

Running with scale factor 1000 results in a file with 186k rows per group, even though the default says 1M

```shell
RUST_LOG=debug nice cargo run --release -- --scale-factor=1000 --format=parquet --tables=lineitem
# check row group size:
andrewlamb@Andrews-MacBook-Pro-3:~/Software/tpchgen-rs$ parquet  meta lineitem.parquet  | grep 'Row group' | head
Row group 0:  count: 183603  44.08 B records  start: 4  total(compressed): 7.718 MB total(uncompressed):12.408 MB
Row group 1:  count: 182366  44.13 B records  start: 8093084  total(compressed): 7.676 MB total(uncompressed):12.330 MB
Row group 2:  count: 183754  44.03 B records  start: 16141473  total(compressed): 7.717 MB total(uncompressed):12.411 MB
...

```


Running with 100 rows per group causes a panic:

```shell
RUST_LOG=debug nice cargo run --release -- --scale-factor=1000 --format=parquet --parquet-row-group-size=100  --tables=lineitem


thread 'tokio-runtime-worker' panicked at tpchgen-cli/src/parquet.rs:111:68:
called `Result::unwrap()` on an `Err` value: General("Parquet does not support more than 32767 row groups per file (currently: 32768)")
```




Instead of breaking up individual parts to limit row group size, what do you think about updating the code to generate parts? I tried to explain how this works in the following PR:
-  https://github.com/clflushopt/tpchgen-rs/pull/157

If we went with the parts approach, we could update this PR to be an adjustment to parts calculation

alamb:
I am going to take the liberty and polish up this PR and get it ready to merge

tpchgen-cli/src/main.rs
```diff
@@ -78,7 +79,7 @@ use tpchgen_arrow::{
 #[command(version)]
 #[command(about = "TPC-H Data Generator", long_about = None)]
 struct Cli {
-    /// Scale factor to address (default: 1)
+    /// Scale factor to create
```
alamb, line 82: I removed the default from the text because clap already puts the default in:

For example:

```shell
  -s, --scale-factor <SCALE_FACTOR>
          Scale factor to create

          [default: 1]
```

tpchgen-cli/src/plan.rs
```diff
@@ -231,21 +249,20 @@ impl OutputSize {
                 Table::Region => 151,
                 Table::Part => 70,
                 Table::Supplier => 164,
-                Table::Partsupp => 140,
+                Table::Partsupp => 141 * 4, // needed to match observed size
```
alamb, line 252: I changed this due when reviewing the output of  the row group size tests and saw the row groups for partsupp were like 4x larger than the target for some reason

tpchgen-cli/src/plan.rs
```diff
@@ -704,6 +721,49 @@ mod tests {
                 "Expected parts {expected_parts:?} should not exceed 32k row groups",
             );
         }
+
+        mod parquet_row_group_size {
```
alamb, line 725: here are the unit tests for this calculation

tpchgen-cli/tests/cli_integration.rs
```diff
@@ -175,6 +176,134 @@ async fn test_write_parquet_orders() {
     }
 }
 
+#[tokio::test]
```
alamb, line 179: I am quite pleased with these tests as they show the overall row group sizes

tpchgen-cli/tests/cli_integration.rs
```diff
@@ -175,6 +176,134 @@ async fn test_write_parquet_orders() {
...
+        output_dir.path(),
+        vec![
+            RowGroups {
+                table: "customer",
+                row_group_bytes: vec![6524595, 6510817, 6509412, 6519647],
+            },
+            RowGroups {
+                table: "lineitem",
+                row_group_bytes: vec![
+                    7159834, 7109252, 7093240, 7123300, 7147731, 7122707, 7144719, 7101681,
```
alamb, line 204: here you can see the row group sizes with scale factor 1 are close to the default target of 7MB

PR #159, opened 2025-07-19
feat: expose answers in tbl format as part of the package

clflushopt:
This pull request exposes TPC-H answers as a module with inlined results in TBL format.


tpchgen/src/q_and_a.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// QueryAndAnswer is a struct that contains a TPC-H query and its expected answer.
+pub struct QueryAndAnswer(
+    &'static str, // The TPC-H query as a string
+    &'static str, // The expected answer as a string
+);
+
+impl QueryAndAnswer {
+    /// Creates a new QueryAndAnswer instance.
+    pub fn new(num: i32) -> Self {
```
alamb, line 15: I recommend adding a `scale_factor: f64` parameter here too (and throwing an error unless it is for scale factor 1) to make the API more explicit abotu what is being returned

alamb: Maybe you could change the signature to the following (return a String as the error)

```suggestion
    pub fn try_new(num: i32) -> Result<Self, String> {
```

tpchgen/src/q_and_a.rs
```diff
@@ -0,0 +1,31 @@
...
+        match num {
+            1..=22 => Self(queries::query(num).unwrap(), answers::answer(num).unwrap()),
+            _ => unreachable!("Invalid TPC-H query number: {}", num),
```
alamb, line 18: it would be nice to avoid a panic here

alamb (approved):
Thanks @clflushopt  -- this seems good to me. Here are some ideas:

## Make the scale factor clear
I think we shoudl make it more clear what scale factor is (I think it is SF1 per the PR description). One way to do this might be:
1. Make a `tpchgen/src/answers/mod.rs` that documents where the data came from
2. Put the answers for scale factor 1 in ``tpchgen/src/answers/sf_1.rs` 

## Document where these queries and answers came from
Could you please document how you generated the answers?
Perhaps in one of these files:
https://github.com/clflushopt/tpchgen-rs/blob/main/README.md
https://github.com/clflushopt/tpchgen-rs/blob/main/TESTING.md

## Suggest a ticket for exposing the answers via the cli
It would be nice to be able to get these answers via `tpchgen-cli` - I suggest filing a ticket to wire up the CLI arguments

## Suggest a test of some kind
It would probably be good to add a basic "is everything hooked up test" -- maybe we could hook the queries/answers up to tpchgen-cli and then use the tpchgen-cli tests in https://github.com/clflushopt/tpchgen-rs/pull/156 🤔

tpchgen/src/q_and_a.rs
clflushopt: I've been pondering both interfaces, `Some(Query, Answer), None` vs marking that as an unreachable state. I'll make the change to use the former.

tpchgen/src/q_and_a.rs
clflushopt: Addressed.

tpchgen/src/q_and_a.rs
clflushopt: Addressed.

clflushopt:
I updated the README to document the source of the answers as well.

clflushopt:
@alamb All comments addressed and created two tracking issues as you suggested would love another review pass when possible.

tpchgen/src/lib.rs
```diff
@@ -51,11 +51,13 @@
 //!
 //! If you want an easy way to generate the TPC-H dataset for usage with external
 //! systems you can use CLI tool instead.
+pub mod answers;
```
alamb, line 54: A minor suggestion might be to consolidate `answers` and `q_and_a`
Maybe something like

```
src/q_and_a.rs             --> src/q_and_a/mod.rs
src/answers/answers_sf1.rs --> src/q_and_a/answers_sf1.rs
src/queries.rs             --> src/q_and_a/queries.rs
```

alamb (approved):
Thanks @clflushopt  -- this looks great

tpchgen/src/lib.rs
clflushopt: Addressed, thanks @alamb !

alamb:
Thanks again @clflushopt

PR #160, opened 2025-07-19
fix: address regression in part count check

clflushopt:
Addresses regression which causes an early exit leading to skipping parallel generation, specifically lines 71 the check is against `1` not `-1`.

Edit: Mistake on my end was looking at the wrong terminal :man_facepalming:

### DIFF
--- a/tpchgen-cli/src/plan.rs
+++ b/tpchgen-cli/src/plan.rs
@@ -68,7 +68,7 @@ impl GenerationPlan {
         cli_part_count: i32,
     ) -> Self {
         // parallel generation disabled if user specifies a part explicitly
-        if cli_part != -1 || cli_part_count != -1 {
+        if cli_part != 1 || cli_part_count != 1 {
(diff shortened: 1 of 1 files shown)


alamb:
> Addresses regression which causes an early exit leading to skipping parallel generation, specifically lines 71 the check is against `1` not `-1`.
> 
> Edit: Mistake on my end was looking at the wrong terminal 🤦‍♂️

BTW I think the use of -1 for the part counts is confusing in general -- I was thinking it would be nicer if we used `Option<i32>` to represent the argument rather than the `-1` value (we can still translate -1 to None during CLI parsing to keep the interface the same, but then we could represent no parts differently internally)

clflushopt:
I agree on both points, in fact I've been thinking about deviating from the hard coded numbers `1, -1` for things like the default part count and no-part count to something more "rustic". Will think about it maybe there's a clean refactor we can make to improve UX.

PR #166, opened 2025-07-29
refactor: Make `--part` and `--parts` `Option<i32>` and add better error handling

alamb:
Follow on to 
- https://github.com/clflushopt/tpchgen-rs/pull/155

I would like to eventually update the meaning of what `--parts` does alone, but to do that reasonably I would like the code in better shape to represent what is going on

# Changes
1. Return an error if only one of `--part` or `--parts` is specified
2. Make `--part` and `--parts` `Option`
3. test error conditions

### DIFF
--- a/tpchgen-cli/src/main.rs
+++ b/tpchgen-cli/src/main.rs
@@ -90,13 +90,15 @@ struct Cli {
     #[arg(short = 'T', long = "tables", value_delimiter = ',', value_parser = TableValueParser)]
     tables: Option<Vec<Table>>,
 
-    /// Number of parts to generate (manual parallel generation)
-    #[arg(short, long, default_value_t = -1)]
-    parts: i32,
+    /// Number of partitions to generate (manual parallel generation)
+    #[arg(short, long)]
+    parts: Option<i32>,
 
-    /// Which part to generate (1-based, only relevant if parts > 1)
-    #[arg(long, default_value_t = -1)]
-    part: i32,
+    /// Which partition to generate (1-based)
+    ///
+    /// If not specified, generates all parts
+    #[arg(long)]
+    part: Option<i32>,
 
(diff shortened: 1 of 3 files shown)


tpchgen-cli/src/main.rs
```diff
@@ -90,13 +90,15 @@ struct Cli {
     #[arg(short = 'T', long = "tables", value_delimiter = ',', value_parser = TableValueParser)]
     tables: Option<Vec<Table>>,
 
-    /// Number of parts to generate (manual parallel generation)
-    #[arg(short, long, default_value_t = -1)]
-    parts: i32,
+    /// Number of partitions to generate (manual parallel generation)
+    #[arg(short, long)]
+    parts: Option<i32>,
```
alamb, line 95: the main purpose of this PR is to make these options. The rest is just error checking and tests

tpchgen-cli/src/plan.rs
```diff
@@ -60,59 +60,98 @@ impl GenerationPlan {
...
-                    part_count: 1,
-                    part_list: vec![1],
-                };
+        match (cli_part, cli_part_count) {
+            (Some(_part), None) => Err(String::from(
+                "The --part option requires the --parts option to be set",
+            )),
+            (None, Some(_part_count)) => {
+                // TODO automatically create multiple files if part_count > 1
+                // and part is not specified
```
alamb, line 79: I also filed a ticket to track this
- https://github.com/clflushopt/tpchgen-rs/issues/167

```suggestion
                // and part is not specified
                // https://github.com/clflushopt/tpchgen-rs/issues/167
```

clflushopt (approved):
LGTM

alamb:
Thank you for the review @clflushopt

PR #169, opened 2025-07-29
refactor: use `RangeInclusive<i32>` in `GenerationPlan`

alamb:
This is a minor refactor to avoid creating Vecs of indices when we can just use a Range.

It likely doesn't really matter, but I hammered it out during a meeting so I figured I would put it up

### DIFF
--- a/tpchgen-cli/src/plan.rs
+++ b/tpchgen-cli/src/plan.rs
@@ -61,59 +61,98 @@ impl GenerationPlan {
     /// Returns the number of parts to generate
     ///
     /// cli_part and cli_part_count are passed from the CLI arguments
-    /// to specify a particular part or number of parts to generate.
-    pub fn new(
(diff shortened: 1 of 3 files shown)


tpchgen-cli/src/plan.rs
```diff
@@ -52,8 +53,8 @@ use tpchgen::generators::{
 pub struct GenerationPlan {
     /// Total number of parts to generate
     part_count: i32,
-    /// List of parts
-    part_list: Vec<i32>,
+    /// List of parts (1..=part_count)
+    part_list: RangeInclusive<i32>,
```
alamb, line 57: The point of the PR is to replace this `Vec`

clflushopt (approved):
Perfect ! LGTM

alamb:
Thank you for the review @clflushopt

Issue #172, opened 2025-08-05
Please make a new PyPI release


MrPowers:
There have been a lot of great PRs merged in recently.  Can you please make a new PyPI release 🙏 

P.S. thank you for making this wonderful project!

MrPowers:
Great, thank you @kevinjqliu!

clflushopt:
We should have 2.0 out soonish PyPI release will be on time with the Rust crates and binaries ! Thanks.

alamb:
I'll try and get the last ticket through

kevinjqliu:
https://pypi.org/project/tpchgen-cli/2.0.0/

<img width="1156" height="536" alt="Image" src="https://github.com/user-attachments/assets/c7227db0-307f-4af4-8cc1-811c1681e022" />

looks like something weird happened with the Project description

kevinjqliu:
https://github.com/clflushopt/tpchgen-rs/pull/182 should fix the README issue on pypi

kevinjqliu:
https://github.com/clflushopt/tpchgen-rs/pull/183 adds a twine check to github action to prevent this from happening again

kevinjqliu:
https://github.com/clflushopt/tpchgen-rs/pull/184 to add `continue-on-error: true` to the github action

PR #184, opened 2025-08-30
infra: add `continue-on-error: true` to the wheel builds

kevinjqliu:
Context: https://github.com/clflushopt/tpchgen-rs/issues/172#issuecomment-3228025491

When any wheel build fails, continue to build the rest of the wheels and publish to pypi

Not 100% sure if we want this behavior...
Worst case scenario, only one wheel is successfully built. The version is released on pypi with 1 wheel. Since pypi artifacts are immutable, we'd have to fix the underlying issue, increase the version, and start a new release

### DIFF
--- a/.github/workflows/tpchgen-cli-publish-pypi.yml
+++ b/.github/workflows/tpchgen-cli-publish-pypi.yml
@@ -11,6 +11,7 @@ permissions:
 jobs:
   linux:
     runs-on: ${{ matrix.platform.runner }}
+    continue-on-error: true
     strategy:
       matrix:
         platform:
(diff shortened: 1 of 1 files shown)


clflushopt:
I think having diverging versions that address diverging builds on different architectures might cause confusion for users, I'd rather we make the version bumps the same across all architectures, even if a minor version update might address just a single architecture issue like we've seen with s390x.

clflushopt:
I agree with @alamb on having a CI test for building the wheels as a separate step that we can do when merging to `main`. Thanks again @kevinjqliu will merge.

PR #187, opened 2025-09-06
CI: ensure python cli wheels and sdist can build

### LINKED ISSUE #186: Add CI test for building wheels
The idea is to ensure we don't accidentally break wheels during development as we saw with s390 in 2.0.0 release:

I agree with @alamb on having a CI test for building the wheels as a separate step that we can do when merging to `main`. Thanks again @kevinjqliu will merge.

_Originally posted by @clflushopt in https://github.com/clflushopt/tpchgen-rs/issues/184#issuecomment-3260141258_

kevinjqliu:
Closes #186

Run python cli build in CI. This will ensure everything passes when we publish to pypi but slows down CI

Refactor to use the same workflow for CI and publishing
Removed `continue-on-error: true` since we want CI to fail when something goes wrong

#### Testing
Added back s390x and verified that CI failed
Pushed change to my fork repo and [manually ran tpchgen-cli-publish-pypi](https://github.com/kevinjqliu/tpchgen-rs/actions/runs/17532338027), verified that all artifacts are built and checked.


kevinjqliu:
Adding back #181 should fail CI

kevinjqliu:
maybe we should publish a new version, tpchgen-cli 2.0.1 since readme was fixed in #182

.github/workflows/build-packages.yml
```diff
@@ -0,0 +1,105 @@
+name: Build Packages
```
clflushopt, line 1:
```suggestion
name: Build Python Packages
```

.github/workflows/rust.yml
```diff
@@ -67,3 +67,9 @@ jobs:
         env:
           RUSTDOCFLAGS: "-D warnings"
         run: cargo doc --no-deps --workspace
+
+  # Build wheels + sdist (but don’t upload artifacts)
+  build-packages:
+    uses: ./.github/workflows/build-packages.yml
```
clflushopt, line 73: I suggest we name the workflow `build-py-packages` to clarify its role.

```suggestion
    uses: ./.github/workflows/build-py-packages.yml
```

clflushopt (approved):
Some minor naming nitpicks and suggestions otherwise LGTM !

.github/workflows/build-packages.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Build wheels
+        uses: PyO3/maturin-action@v1
+        with:
+          target: ${{ matrix.target }}
+          args: --release --out dist
+          working-directory: tpchgen-cli
+          manylinux: ${{ matrix.manylinux || '' }}
+      - name: Upload wheels
+        if: ${{ inputs.upload-artifacts }}
+        uses: actions/upload-artifact@v4
```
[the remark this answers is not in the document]
kevinjqliu: yea the `actions/upload-artifact` action "publishes" the artifacts on github, which just means the artifacts show up in the specific github action run. 
For example, https://github.com/kevinjqliu/tpchgen-rs/actions/runs/17518754727
the wheels and sdist are in the "Artifacts" section 

In `.github/workflows/tpchgen-cli-publish-pypi.yml`, these artifacts are downloaded to the github runner's env before uploading to pypi
```
      - name: Download all build artifacts
        uses: actions/download-artifact@v4
        with:
          pattern: '*'
          merge-multiple: true
          path: dist
```

kevinjqliu:
Feel free to merge this, we can triple check the `tpchgen-cli-publish-pypi.yml` workflow by running it in this repo after merging the PR. 
Since we use `--non-interactive --skip-existing dist/*` for the upload command, rerunning the action should be a no-op

We can also bump to 2.0.1 and republish to fix the README, https://pypi.org/project/tpchgen-cli/2.0.0/#description

alamb:
Let's try it!

alamb:
Thanks @kevinjqliu  and @clflushopt

PR #191, opened 2025-11-01
infra:  configure dependabot for cargo & github action updates

pmcgleenon:
I've added dependabot configuration for github actions and cargo dependencies since we are using a bunch of external dependencies and some of them have updates available (e.g. `actions/checkout@v4`).


clflushopt:
Thanks @pmcgleenon this was an omission on my part

Issue #51, opened 2025-03-20
[FEATURE] Extend to support TPC-DS data generation


scsmithr:
**Is your feature request related to a problem? Please describe.**

I think it'd be cool to have this be able to generate TPCH-DS data as well.

Opening more for discussion. I've spent significantly less time looking the TPC-DS spec (or related gen code), all I know is there's a lot more tables to generate.

**Describe the solution you'd like**

Generate TPC-DS data

**Describe alternatives you've considered**

**Additional context**

clflushopt:
Thanks for opening this issue, I had this in the back of my mind but kind of forgot about :), I did look at TPC-DS spec and I also spent sometime looking at the `dsdgen` implementation so I have an idea about how to write up a first port; unfortunately I won't be able to make it in the first milestone we have `v0.1.0` but it will be the first item I'll look at after we cut the first release.

alamb:
FWIW @sundy-li asked about this on twitter: https://x.com/sundyli1/status/1910875165145051390

I don't have any plans to work on TPC-DS myself but there might be community interest

sundy-li:
Actually we are using duckdb to generate tpch/tpcds data in  [bendsql](https://github.com/databendlabs/bendsql/blob/c66816de8824269ebb54cbfce47d6060b251aa97/cli/src/gendata.rs#L66-L76) , it'll be awesome if this tool works for tpcds data.

clflushopt:
I have plans to address this. I already started looking at the existing `dsdgen` and I am evaluating how much time & lift it would take.

clflushopt:
I did !  I am still looking at the existing implementation in C++ [here](https://github.com/askyx/pg_tpcds/blob/main/src/tpcds/dsdgen/w_item.cpp) and will start tinkering about the port soonish

alamb:
Thank you @clflushopt  -- I am sorry I haven't been around / active as much related to TPCH recently. Things have been quite busy in DataFusion / Arrow land.

One thing I had been thinking of trying was one of the new AI coding assistance / agents -- like perhaps setting up the project and porting one table. Then we can turn one of the many tools loose on following the same pattern and porting the other tables 🤔

djouallah:
Any update on this , btw the latest tpch is so fast I thought there was a bug, what an awesome job

clflushopt:
Hey @djouallah no update for now, I haven't started drafting a plan for this yet, but it's on the top of my list of things to work on next.

[...]

PR #212, opened 2026-01-06
feat: TPC-DS support

clflushopt:
This PR adds TPC-DS support which was built initially out of tree in [tpcdsgen](https://github.com/clflushopt/tpcdsgen)

### DIFF
--- a/.github/workflows/tpcdsgen-conformance.yml
+++ b/.github/workflows/tpcdsgen-conformance.yml
@@ -11,50 +11,10 @@ env:
   RUST_BACKTRACE: 1
 
 jobs:
-  # Job 1: Build and test Rust code
-  rust-tests:
-    name: Rust Tests
-    runs-on: ubuntu-latest
-
-    steps:
-    - name: Checkout repository
-      uses: actions/checkout@v4
-
-    - name: Install Rust toolchain
-      uses: dtolnay/rust-toolchain@stable
-      with:
-        components: rustfmt, clippy
-
-    - name: Cache Rust dependencies
-      uses: actions/cache@v4
-      with:
-        path: |
-          ~/.cargo/bin/
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-        restore-keys: |
-          ${{ runner.os }}-cargo-
-
-    - name: Check formatting
-      run: cargo fmt -p tpcdsgen -- --check
-
-    - name: Run clippy
-      run: cargo clippy -p tpcdsgen -- -D warnings
-
-    - name: Build
-      run: cargo build -p tpcdsgen --verbose
-
-    - name: Run unit tests
-      run: cargo test -p tpcdsgen --verbose
-
-  # Job 2: Conformance testing against Java implementation
+  # Conformance testing against Java implementation
   conformance-tests:
     name: Conformance Tests
     runs-on: ubuntu-latest
-    needs: rust-tests  # Only run if Rust tests pass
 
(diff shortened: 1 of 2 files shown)


alamb:
This is on my list to review -- hopefully I'll have time to do it later today

.github/workflows/tpcdsgen-conformance.yml
```diff
@@ -0,0 +1,154 @@
...
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-cargo-
+
+    - name: Check formatting
```
alamb: These seem unrelated to conformance testing -- they are more like just normal rust code checks -- I recommend we put them in a different job file (can be done as a follow on PR)

.github/workflows/tpcdsgen-conformance.yml
```diff
@@ -0,0 +1,154 @@
...
+    - name: Run clippy
+      run: cargo clippy -p tpcdsgen -- -D warnings
+
+    - name: Build
+      run: cargo build -p tpcdsgen --verbose
+
+    - name: Run unit tests
+      run: cargo test -p tpcdsgen --verbose
+
+  # Job 2: Conformance testing against Java implementation
```
alamb: I thought about this -- one thing we might consider doing is only check in the java md5 hash of the output (which doesn't change)  which would make the tests run much faster (no need to build and run java)

Also something potentially for a follow on PR

tpcdsgen/scripts/README.md
```diff
@@ -0,0 +1,341 @@
+# TPC-DS Test Scripts
+
+This directory contains scripts for testing the Rust TPC-DS implementation against the Java reference implementation.
```
alamb: ❤️

alamb (approved):
👏  pretty epic @clflushopt  --

I took a careful look at the github stuff and gave a cursory review of the other 220 files (!!) and I think we should proceed with merging it and file some follow on tasks

The code I read was well structured, well commented, and I think will be straightforward to maintain / update as needed. Well done. 


Some ideas for follow on tasks
1. Work on conformance testing with C (e.g. https://github.com/clflushopt/tpchgen-rs/issues/51#issuecomment-3700913731)
2. Improve CI checks (e.g. figure out how to avoid having to maintain all the java setup, etc)
3. start integrating into the cli

alamb:
I noticed several comments about "bugs in the C code" which I assume come from the Java port. It wasn't clear to me of the java people "fixed" the bugs in the C code -- if they did this could be the source of the differences between the C implementation and the rust one

I would personally prefer to follow the C semantics (bugs and all) as that is the 'source of truth'  but that can be a discussion we have after his is merged

Thank you again -- this is pretty epic

clflushopt:
> I noticed several comments about "bugs in the C code" which I assume come from the Java port. It wasn't clear to me of the java people "fixed" the bugs in the C code -- if they did this could be the source of the differences between the C implementation and the rust one
> 
> I would personally prefer to follow the C semantics (bugs and all) as that is the 'source of truth' but that can be a discussion we have after his is merged
> 
> Thank you again -- this is pretty epic

Actually the documented bugs are not fixed in the Java version and are not fixed in the Rust implementation as well they are just documented.
My hunch tells me that there are more bugs in the way the RNG streams synchronization logic works in the C implementation that were fixed in the Java version "accidently" maybe because of language/runtime differences (probably an overflow/underflow or some signedness non-sense). In any case I will get to the root cause and fix it.

Meanwhile thanks for the review, I will address the items for CI in the same PR (it's easier this way).

PR #213, opened 2026-01-07
docs: fix typo in architecture guide

Matt711:
I can also help add some sort of hook to CI so we can catch these kinds of typos earlier. Did you have any tool suggestions?


clflushopt (approved):
LGTM

kevinjqliu:
> Ideally I'd rather not add a new dependency in the CI but I am open to suggestions

iceberg-rust has typo check as running in CI, https://github.com/apache/iceberg-rust/blob/76cdf280231bf27ea888de15507d918392f08d44/.github/workflows/ci_typos.yml#L18

but i think having the ability to check locally is better.

Issue #215, opened 2026-01-09
[FEATURE] Support custom delimiters when CSV is chosen as the output format.


WSL0809:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when importing generated CSV files into databases/tools that require non-comma delimiters (e.g., ;, \t), as the current implementation hardcodes the comma separator and forces extra post-processing steps.

**Describe the solution you'd like**

Add a command-line option (e.g., --csv-delimiter <CHAR>) to allow specifying a custom CSV field separator, with comma as the default value for backward compatibility.

**Describe alternatives you've considered**

Use sed/awk to batch replace commas with the target delimiter after generation, or switch to TBL format (pipe-separated). Both add extra steps and reduce workflow efficiency.

clflushopt:
I have a tentative PR for this see https://github.com/clflushopt/tpchgen-rs/pull/219

PR #219, opened 2026-01-11
feat: custom CSV delimiter support

clflushopt:
- closes https://github.com/clflushopt/tpchgen-rs/issues/215

### DIFF
--- a/tpchgen-cli/bin/main.rs
+++ b/tpchgen-cli/bin/main.rs
@@ -124,9 +124,33 @@ struct Cli {
     /// Specifies the delimiter character to use when generating CSV files.
     /// This option only applies to CSV format and cannot be used with TBL format.
     ///
+    /// Supports escape sequences: \t (tab), \n (newline), \r (carriage return), \\ (backslash)
     /// Common delimiters: ',' (comma), '|' (pipe), '\t' (tab), ';' (semicolon)
-    #[arg(long, default_value_t = ',')]
+    #[arg(long, default_value = ",", value_parser = parse_delimiter)]
     delimiter: char,
+}
+
+/// Parse a delimiter string, handling escape sequences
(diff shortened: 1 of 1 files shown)


alamb:
FYI @WSL0809

clflushopt:
I added some logic in the last commit to deal with delimiters that are multiple characters but we restrict it to a subset of escape sequences.

The rule is that :
- Delimiter is a single character.
- Delimiter is multi character in this case it can only be one of the escape sequences we accept.
- Delimiter is specified when `--format=tbl` is passed will reject it so we don't accidentally break from TBL format.

kevinjqliu (approved):
LGTM

Tried the 4 supported escape sequence, and an unsupported one
```
cargo run --bin tpchgen-cli --release -- --tables region --format csv --delimiter '\t'
```

tpchgen/src/csv.rs
```diff
@@ -407,24 +512,39 @@ impl Display for OrderCsv<'_> {
...
+        Self { inner, delimiter }
     }
 
     /// Returns the CSV header for the LineItem table
     pub fn header() -> &'static str {
         "l_orderkey,l_partkey,l_suppkey,l_linenumber,l_quantity,l_extendedprice,l_discount,l_tax,l_returnflag,l_linestatus,l_shipdate,l_commitdate,l_receiptdate,l_shipinstruct,l_shipmode,l_comment"
     }
+
+    /// Returns the CSV header with a custom delimiter
+    pub fn header_with_delimiter(delimiter: char) -> String {
```
alamb: A minor suggestion is you could avoid the repetition by changing `header` to call `header_with_delimiter` -- something like
```rust
    pub fn header() -> &'static str {
      Self::header_with_delimiter("")
    }
```

PR #225, opened 2026-01-13
cli: add `--quiet` flag, enable logging by default

kevinjqliu:
Great suggestion from @alamb to add a `--quiet` flag that suppresses all output, https://github.com/clflushopt/tpchgen-rs/pull/220#pullrequestreview-3651624336 
`--quiet` sets the log level to `ERROR` so the CLI can still display error logs 

I also took the opportunity to refactor the remaining usages of `eprintln!` and `println!`, replacing them with log statements. 

We still retain the behavior implemented in #220; the cli still warns when skip generation
Screenshot 2026-01-13 at 2 20 35 PM

PR #226, opened 2026-01-14
feat: Track table generation with progress bars

### LINKED ISSUE #216: [FEATURE] Track table generation with progress bars
This is especially helpful when generating larger datasets. This feature will help users tell how much progress has been made per table. And it's something to look at while they wait! 😅 

Solution: Add an optional progress indicator to `tpchgen-cli` that displays progress bars while data is being generated.

I'm still reading the up on the the two main crates `tpchgen` and `tpchgen-cli`, so I have less specifics in this section, but I think all changes should be scoped to tpchgen-cli crate given the note about [dependencies](https://github.com/clflushopt/tpchgen-rs/blob/main/ARCHITECTURE.md#dependencies).

Ex: Just the idea
```bash
$ tpchgen-cli ... --track or --progress
```
```bash
orders     [xxxxxxxxxxxxxx                          ] 50%
lineitem   [x                                       ] 10%  
...
```

At a high level,
- Initialize a progress tracker when a `--track` flag is passed
- Hook into existing data generation flow and update the tracker. Somewhere https://github.com/clflushopt/tpchgen-rs/blob/d16d958244fc5937c2481cf38451e5e254404211/tpchgen-cli/src/generate.rs#L51 I think?
- Mark tables complete when generation finishes

None, open

**Additional context**

<details>

<summary>For tpchgen-rs, with ❤️ :</summary>

Massive thanks for the work you’ve put into tpchgen-rs. I’ve been spending time reading through the repo to learn more Rust, and the code has been a pleasure to study.

I’m especially looking forward to the upcoming TPC-DS data generator (another gigantic ❤️ @clflushopt) that’s a big part of what motivated me to dig into the project now.

For the terminal display, I have been reading about using [indicatif](https://docs.rs/indicatif/latest/indicatif/). It is written in Rust and I've been getting more familiar with their API. Right now it's my choice, but open to suggestions here too.

</details>

Matt711:
Closes #216

Adds progress reporting for table generation, including default CLI progress bars and an optional library progress hook.

## Design

Generation reports progress through a shared, thread-safe `ProgressTracker`. The CLI default implementation, `IndicatifProgress`, uses `indicatif` with one progress bar per `Table`, coordinated by `MultiProgress`.

The library-facing API stays small:

- `ProgressTracker` is the public hook for custom progress sinks.
- `RunProgress` tracks run-level totals, skipped files, and finish behavior.
- `TableProgress` is passed to writers and advances progress for completed output units.

Progress is advanced only after output is successfully written. Since generation parallelism is across tables, each table writer still emits its chunks or row groups in order, so each table bar moves forward one output unit at a time.

Progress units are format-specific:

- TBL/CSV counts written data chunks, excluding headers.
- Parquet counts written row groups.
- Skipped existing files advance by their planned output-unit count.

## CLI Behavior

Progress bars are enabled by default only when stderr is an interactive terminal. They are suppressed by `--no-progress`, `--quiet`, `--stdout`, or non-TTY stderr, keeping CI logs and piped output clean.

Library users can enable the lightweight `progress` feature and provide their own `ProgressTracker` without depending on the default terminal progress implementation.

https://github.com/user-attachments/assets/ca7c0370-47e8-489a-aa85-29b4a673c7d7

### DIFF
--- a/tpchgen-cli/README.md
+++ b/tpchgen-cli/README.md
@@ -72,6 +72,11 @@ for PART in `seq 2 3`; do
 done
 ```
 
+By default `tpchgen-cli` shows a per-table progress bar on stderr while data
+is generated. Pass `--no-progress` to disable it (it is also disabled
+automatically when `--quiet` is set, when `--stdout` is used, or when stderr
+is not a terminal, e.g. in CI logs).
+
 ## Performance
 
 | Scale Factor | `tpchgen-cli` | DuckDB     | DuckDB (proprietary) |
--- a/tpchgen-cli/bin/main.rs
+++ b/tpchgen-cli/bin/main.rs
@@ -10,14 +10,10 @@
 use clap::builder::TypedValueParser;
 use clap::{ArgAction, Parser};
 use log::{info, LevelFilter};
-use std::io;
-#[cfg(feature = "indicatif-progress")]
-use std::io::IsTerminal;
+use std::io::{self, IsTerminal};
 use std::path::PathBuf;
 use std::str::FromStr;
-#[cfg(feature = "indicatif-progress")]
 use std::sync::Arc;
-#[cfg(feature = "indicatif-progress")]
 use tpchgen_cli::progress::IndicatifProgress;
 use tpchgen_cli::{
     Compression, OutputFormat, Table, TpchGenerator, TpchGeneratorBuilder,
@@ -138,7 +134,6 @@ impl CommonArgs {
     /// Initialize CLI logging/progress output and create a
     /// [`TpchGeneratorBuilder`] pre-configured with the common options.
     fn builder(self, format: OutputFormat) -> TpchGeneratorBuilder {
-        #[cfg(feature = "indicatif-progress")]
         let progress = self
             .should_show_progress_bars()
             .then(|| Arc::new(IndicatifProgress::new()));
@@ -160,24 +155,18 @@ impl CommonArgs {
             builder = builder.with_part(part);
         }
 
-        #[cfg(feature = "indicatif-progress")]
-        {
-            configure_logging(
-                self.verbose,
-                self.quiet,
-                progress.as_ref().map(|progress| progress.log_writer()),
-            );
-            if let Some(progress) = progress {
-                builder = builder.with_progress_tracker(progress);
-            }
+        configure_logging(
+            self.verbose,
+            self.quiet,
+            progress.as_ref().map(|progress| progress.log_writer()),
+        );
+        if let Some(progress) = progress {
+            builder = builder.with_progress_tracker(progress);
         }
-        #[cfg(not(feature = "indicatif-progress"))]
-        configure_logging(self.verbose, self.quiet, None);
 
         builder
     }
 
-    #[cfg(feature = "indicatif-progress")]
     fn should_show_progress_bars(&self) -> bool {
         // Show progress only on an interactive terminal and when no flag
         // suppresses it. `--stdout` is included so piped data isn't
--- a/tpchgen-cli/bin/main.rs
+++ b/tpchgen-cli/bin/main.rs
@@ -8,7 +8,7 @@
 
 // Use the library public API
 use clap::builder::TypedValueParser;
-use clap::Parser;
+use clap::{ArgAction, Parser};
 use log::{info, LevelFilter};
 use std::io;
 use std::path::PathBuf;
@@ -120,8 +120,8 @@ struct CommonArgs {
     #[arg(long, default_value_t = false)]
     stdout: bool,
 
-    /// Show progress bars during data generation (shown by default, hidden with --quiet)
-    #[arg(short = 'P', long, default_value_t = true)]
+    /// Disable progress bars during data generation
+    #[arg(long = "no-progress", action = ArgAction::SetFalse, default_value_t = true)]
     progress: bool,
 }
 
--- a/tpchgen-cli/bin/main.rs
+++ b/tpchgen-cli/bin/main.rs
@@ -11,8 +11,14 @@ use clap::builder::TypedValueParser;
 use clap::{ArgAction, Parser};
 use log::{info, LevelFilter};
 use std::io;
+#[cfg(feature = "indicatif-progress")]
+use std::io::IsTerminal;
 use std::path::PathBuf;
 use std::str::FromStr;
+#[cfg(feature = "indicatif-progress")]
+use std::sync::Arc;
+#[cfg(feature = "indicatif-progress")]
+use tpchgen_cli::progress::IndicatifProgress;
 use tpchgen_cli::{
     Compression, OutputFormat, Table, TpchGenerator, TpchGeneratorBuilder,
     DEFAULT_PARQUET_ROW_GROUP_BYTES,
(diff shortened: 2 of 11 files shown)


alamb:
Thank you @Matt711  -- this looks very cool. I am not sure I will have time to review it properly. The screen recording looks really nice

One question I had was what do you think about showing progress by default? It seems like this might interact weirdly with https://github.com/clflushopt/tpchgen-rs/pull/225 from @kevinjqliu -- specifically logging)

Matt711:
> Just merged #225 which defines the logging level based on the given param (default, `--quiet`, `--verbose`)
> 
> https://github.com/clflushopt/tpchgen-rs/blob/b91de80b403e1b094bbf8f5ea5059e5da0feef9f/tpchgen-cli/bin/main.rs#L177-L192
> 
> I think we should take this into consideration when rendering the process bar. i.e. turn progress bar off for `--quiet`

Done. We now turn off the progress bar for `--quiet`. We also log when files are skipped.

clflushopt (approved):
This LGTM I'll wait for another review/resolve pass from @alamb before merging.

clflushopt (approved):
LGTM ! Thanks @Matt711 awesome stuff !

kevinjqliu:
i'll rebase and move this forward

tpchgen-cli/src/lib.rs
```diff
@@ -541,3 +573,61 @@ impl Default for TpchGeneratorBuilder {
         Self::new()
     }
 }
+
+#[cfg(all(test, feature = "progress"))]
```
kevinjqliu, line 580: adds a test for custom progress tracker

tpchgen-cli/bin/main.rs
```diff
@@ -394,20 +429,28 @@ impl ParquetArgs {
...
     } else {
         // Default: show warnings and errors, but respect RUST_LOG if set
-        env_logger::builder()
-            .filter_level(LevelFilter::Warn)
-            .parse_default_env()
-            .init();
+        builder.filter_level(LevelFilter::Warn).parse_default_env();
+    }
+    if let Some(log_writer) = log_writer {
+        builder.target(env_logger::Target::Pipe(log_writer));
```
kevinjqliu, line 437: this solves the issue when both progress tracker and log needs to write to the same output

tpchgen-cli/bin/main.rs
```diff
@@ -8,11 +8,17 @@
 
 // Use the library public API
 use clap::builder::TypedValueParser;
-use clap::Parser;
+use clap::{ArgAction, Parser};
 use log::{info, LevelFilter};
 use std::io;
+#[cfg(feature = "indicatif-progress")]
```
kevinjqliu: cli can be compiled without `indicatif`, but does not allow custom progress tracker.

tpchgen-cli/src/generate.rs
```diff
@@ -109,10 +168,13 @@ where
     // The writer task runs in a blocking thread to avoid blocking the async
     // runtime. It reads from the channel and writes to the sink (doing File IO)
     let captured_recycler = recycler.clone();
+    let captured_progress = progress.clone();
     let writer_task = tokio::task::spawn_blocking(move || {
+        sink.sink(&header)?;
```
kevinjqliu, line 141: Write the header first, then send data chunks through the channel, so progress counts every channel item without needing to skip the first one.

kevinjqliu (commented):
i kind of went over board with codex 😆 i was trying to do some refactor so that it progress tracker is as isolated as possible
im happy with the code now. @alamb @clflushopt if you would like to take a look

kevinjqliu:
btw big thank you to @Matt711 for the implementation

alamb:
I checked this adds about 300k to the binary size

This branch:
```shell
alamb@aal-dev:~/tpchgen-rs$ du -s -h ../tpchgen-rs2/target/release/tpchgen-cli
9.6M    ../tpchgen-rs2/target/release/tpchgen-cli
```

main:
```
alamb@aal-dev:~/tpchgen-rs$ du -s -h ./target/release/tpchgen-cli
9.9M    ./target/release/tpchgen-cli
```

I also tried it locally and it works really nicely 👌 
status2

tpchgen-cli/bin/main.rs
alamb: I would personally lean towards making this not optional. If people want a simpler CLI they can write it themselves from the lower level crates (aka I suggest removing the feature flag)

tpchgen-cli/src/generate.rs
```diff
@@ -49,9 +50,38 @@ pub trait Sink: Send {
...
+    sources: I,
+    num_threads: usize,
+    progress: TableProgress,
+) -> Result<(), io::Error>
+where
+    G: Source + 'static,
+    I: Iterator<Item = G>,
+    S: Sink + 'static,
+{
+    generate_in_chunks_impl(sink, sources, num_threads, progress).await
```
alamb, line 77: I recommend  simplifying this by just adding progres to `generate_in_chunks` perhaps (and avoiding the second signature()

tpchgen-cli/src/generate.rs
alamb: maybe worth a comment about that

tpchgen-cli/src/parquet.rs
```diff
@@ -33,6 +34,47 @@ pub async fn generate_parquet<W: Write + Send + IntoSize + 'static, I>(
...
+        writer,
+        iter_iter,
+        num_threads,
+        parquet_compression,
+        progress,
+    )
+    .await
+}
+
+pub(crate) async fn generate_parquet_with_progress<W: Write + Send + IntoSize + 'static, I>(
```
alamb, line 51: same comments as above about how to simplify this code

tpchgen-cli/src/progress.rs
```diff
@@ -0,0 +1,442 @@
...
+use crate::output_plan::OutputPlan;
+use crate::Table;
+#[cfg(feature = "progress")]
+use std::collections::BTreeMap;
+#[cfg(feature = "progress")]
+use std::fmt;
+#[cfg(feature = "progress")]
+use std::sync::Arc;
+
+/// Receives generation-progress events for one
```
alamb, line 72: this is very nice

alamb (approved):
Looks great to me -- thank you @kevinjqliu and @Matt711 

I had some simplifying suggestions, but we could do them as a follow on too if we want

One of the major ones is to remove the feature flag as I think it isn't super helpful

tpchgen-cli/README.md
```diff
@@ -72,6 +72,11 @@ for PART in `seq 2 3`; do
 done
 ```
 
+By default `tpchgen-cli` shows a per-table progress bar on stderr while data
+is generated. Pass `--no-progress` to disable it (it is also disabled
```
alamb, line 76: I wonder if we really need --no-progress AND --quiet? Maybe `--quiet` is enough?

tpchgen-cli/src/generate.rs
kevinjqliu: I did this so that there's still backwards compatibility for `generate_parquet`, when users do not want to deal with using the optional `progress` feature. 
`generate_parquet_with_progress` we call internally and `generate_parquet_impl` is the shared implementation

i found this to be cleaner than adding `progress` arguments to the public API, happy to take suggestions!

tpchgen-cli/README.md
kevinjqliu: Right now `--quiet` turns off the progress tracker and suppress info/warn logs. If we get rid of `--no-progress`, users will not have a way to keep info/warn logs while keeping the progress tracker disabled... 

That’s a narrow use case, so I’m happy to follow up and remove `--no-progress` if you feel strongly about it.

kevinjqliu:
Made these changes:
* Made `indicatif-progress` required for CLI. Thats a good call, the `main.rs` code looks much cleaner now.
* Added a comment about sinking the header

Responded about `--no-progress` and `generate_in_chunks`, I can follow up with PRs if needed

kevinjqliu:
I have to mark all conversation as resolved to see the merge button.

kevinjqliu:
Lets merge this! 🚢 🚢 🚢 

Thanks @Matt711, @clflushopt and @alamb

alamb:
this next release is going to be so good

Issue #214, opened 2026-01-08
investigate checking for typos


kevinjqliu:
Context in https://github.com/clflushopt/tpchgen-rs/pull/213

Either in CI or in local env or both

iceberg-rust has typo check as running in CI, https://github.com/apache/iceberg-rust/blob/76cdf280231bf27ea888de15507d918392f08d44/.github/workflows/ci_typos.yml#L18

Matt711:
I punted on CI in #217, but am happy to vendor what iceberg-rust does over here if that's what we decide to do

PR #227, opened 2026-01-14
Add typos to CI

Matt711:
One more file type exclusion after #212 

- xref #214

### DIFF
--- a/.github/workflows/ci_typos.yml
+++ b/.github/workflows/ci_typos.yml
@@ -1,4 +1,4 @@
-# Copyright The Apache Software Foundation
+# Copyright The tpchgen-rs Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


.github/workflows/ci_typos.yml
```diff
@@ -0,0 +1,45 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
kevinjqliu: this is not an apache repo, so i dont think we should include the ASF license 😄 

but feel free to add a comment and mention its copied from https://github.com/apache/iceberg-rust/blob/76cdf280231bf27ea888de15507d918392f08d44/.github/workflows/ci_typos.yml#L18

.github/workflows/ci_typos.yml
```diff
@@ -0,0 +1,42 @@
+# Copyright The Apache Software Foundation
```
kevinjqliu: @alamb could you take a look at this license part?

kevinjqliu: thanks for catching this! that makes sense. 
this is documented in https://www.apache.org/licenses/LICENSE-2.0#apply

kevinjqliu:
```suggestion
# Copyright The tpchgen-rs Authors
```

kevinjqliu:
Thanks @Matt711 and thank you @alamb for the review!

PR #253, opened 2026-05-10
feat: dsdgen compatibility mode (with `C`)

clflushopt:
## Add C dsdgen compatibility mode and conformance tooling

This PR adds a `--compat` flag to `tpcdsgen`, fixes several inherited bugs from the Java port, and introduces a comparison script for validating output against the C dsdgen reference implementation.

### Background

TPC-DS data generation has two reference implementations: the original C `dsdgen` and the Java port used by Trino. Our Rust implementation was ported from the Java side, inheriting its behavior. This PR makes both behaviors available and correct i.e. we reproduce bugs from both versions.

We introduce a top-level `--compat trino|c` flag (default: `trino`) plumbed through the full config via a new `CompatMode` enum. The default preserves byte-for-byte compatibility with the Java/Trino implementation. Passing `--compat c` corrects known divergences to match C dsdgen (we can go the other way, defaulting to C compatibility instead).

### "Bugs fixed" I mean "behaviour replicated" under `--compat c`

**1. `reason` table: wrong row count at all scale factors**

The Java port encoded `reason` as a logarithmic-scaling table, producing 35 rows at SF1 but C dsdgen generates all 75 entries from `return_reasons.dst` for each scale factor. Under `--compat c`, `Scaling::get_row_count` returns the correct static count of 75.

**2. `*_reason_sk` foreign keys in returns tables**

`catalog_returns`, `store_returns`, and `web_returns` generate reason FKs by sampling uniformly from `1..=reason_count`. With the wrong count of 35, roughly 90% of FK values in each table diverged from C dsdgen this is fixed automatically by the reason row count fix above.

**3. `return_reasons.dst`: missing `reason 30` entry**

The inherited `.dst` file had `reason 31` duplicated at positions 30 and 31, with `reason 30` absent entirely. A corrected `return_reasons_c.dst` is introduced for the C compat path; the original file is kept for Trino compat, since the Java implementation carries the same data and its fixtures must remain byte-for-byte stable.

### C dsdgen conformance tooling

Initially I introduced `datafusion-benchmarks` as a git submodule but the repo structure got all messed up and trying to get a fix for the "Latin-1" issue as a commit in the submodule didn't workout quite as expected.

**`scripts/compare-c.py`** — a new script that compares `tpcdsgen --compat c` output against C dsdgen Parquet reference data:

- Reads `.dat` files as Latin-1 (matching both Rust and C dsdgen output encoding)
- Normalizes Parquet string values to handle mojibake from older reference data
- Sorts both sides canonically before diffing to handle ordering differences
- Reports row-count mismatches, diff percentages, and column-level diffs with `--verbose`
- Takes `--dat-dir` and `--parquet-dir` as explicit arguments — no hard-coded paths or submodule dependency

**`scripts/tpcds_schemas.py`** — standalone PyArrow schema definitions for all 24 TPC-DS tables, extracted from `apache/datafusion-benchmarks` (Apache-2.0). So no submodule needed.

### Scripts cleanup

- Replaced `ls` glob patterns with `find` for reliable JAR discovery in `bootstrap-java.sh` and `generate-fixtures.sh`
- Fixed POSIX-unsafe arithmetic in `generate-fixtures.sh`
- Hardened `benchmark.sh` with `set -euo pipefail`
- Corrected stale content in `scripts/README.md`

### Conformance results

| Mode | Result |
|---|---|
| `--compat trino` (default) | 24/24 tables pass byte-for-byte against Java reference fixtures |
| `--compat c` | 24/24 tables match C dsdgen Parquet reference at SF1 |


tpcdsgen/scripts/compare-c.py
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env python3
```
alamb: It would be nice to consolidate the the testing / comparison for java and C

it seems like the code to compare with c is now done in python but the scripts to compare with java is done with .sh. Both are fine, but it is strange to me that they are different

```
andrewlamb@Andrews-MacBook-Pro-3:~/Software/tpchgen-rs$ ls tpcdsgen/scripts/
benchmark.sh*         clean-fixtures.sh*    compare-table.sh*     README.md             tpcds_schemas.py
bootstrap-java.sh*    compare-c.py          generate-fixtures.sh* test-all-tables.sh*
```

tpcdsgen/scripts/compare-c.py
```diff
@@ -0,0 +1,227 @@
...
+"""
+Compare Rust .dat output against C dsdgen Parquet reference.
+
+Typical workflow:
+    # 1. Obtain C dsdgen Parquet reference data, e.g. from datafusion-benchmarks:
```
alamb: I think as a follow on PR it would be good to change this verification to use the actual dbdsgen.c data generator (maybe we can pre-compute the results and check it into some repository) rather than use the parquet version -- if we go through parquet we have a potential level of indirection / modification that may mess up things

tpcdsgen/src/config/scaling.rs
```diff
@@ -1,22 +1,39 @@
-use crate::config::Table;
+use crate::config::{CompatMode, Table};
 use crate::distribution::calendar_distribution::{CalendarDistribution, CalendarWeights};
 use crate::table::Table as MetaTable;
 use crate::types::Date;
 
+/// Row count for the `reason` table under C dsdgen compatibility.
+///
+/// C dsdgen generates all 75 entries from `return_reasons.dst` at every scale
+/// factor. The Java/Trino port incorrectly used a logarithmic scaling model
```
alamb: I wonder if there is a reference / ticket we can refer to here (eg. a github link to the java source ) so it is clearer how we know this

tpcdsgen/src/distribution/return_reasons_distribution.rs
```diff
@@ -19,12 +19,26 @@ impl ReturnReasonsDistribution {
         })
     }
 
-    /// Get return reason at the specified index (getValueAtIndex)
-    pub fn get_return_reason_at_index(index: usize) -> Result<&'static str> {
-        Self::get_distribution().get_value_at_index(0, index)
+    /// C dsdgen uses a corrected .dst where `reason 30` is not a duplicate of `reason 31`.
```
alamb: same thing here -- a link to the reference / details about this divergence I think would be super helpful

tpcdsgen/main.py
```diff
@@ -0,0 +1,6 @@
+def main():
```
alamb: this seems unecessary 🤔

tpcdsgen/uv.lock
alamb, line 1: do we need to check in the uv.lock fiel?

alamb (approved):
This looks great - thank you @clflushopt 

I think there are a few more things I recommend we do as follow on PRs:
1. Pull out the CLI generation into its own crate (tpchdsgen-cli perhaps) similarly to tpchgen-cli
2. Create a tpchdsgen Arrow crate
3. Rework the verification scripts so they don't have to use parquet (they can use pre-computed tpcds data files)
I will file tickets to track these

alamb:
Typos check looks readl: https://github.com/clflushopt/tpchgen-rs/actions/runs/25615843698/job/75193615484?pr=253

alamb:
I got a little carried away this morning and made a new repo with pre-generated tpcds data: https://github.com/alamb/tpcds-data#

I'll work on a PR to your branch that uses this pre-computed data to compare again ratther than the parquet files in datafusion-benchmarks

tpcdsgen/scripts/compare-c.py
alamb: I am having good luck consolidating scripts in the following PR (a proposed update to this one)
- https://github.com/clflushopt/tpchgen-rs/pull/257

tpcdsgen/uv.lock
alamb: removed

tpcdsgen/scripts/bootstrap-trino.sh
```diff
@@ -49,32 +80,6 @@ log_error() {
     echo -e "${RED}[ERROR]${NC} $*" >&2
 }
 
-# Print usage
-usage() {
```
alamb, line 53: moved this up and consolidated the descriptions from readme and comments into a single place

I did this for all of the scripts to reduce the amount of redundancy

.github/workflows/tpcdsgen-conformance.yml
```diff
@@ -45,7 +51,7 @@ jobs:
     - name: Bootstrap Java TPC-DS implementation
       run: |
         cd tpcdsgen
-        ./scripts/bootstrap-java.sh
+        ./scripts/bootstrap-trino.sh
```
alamb, line 54: I also renamed "java" to "trino" so that it was consistent across the codebase

tpcdsgen/scripts/README.md
```diff
@@ -1,285 +1,143 @@
 # TPC-DS Test Scripts
 
-This directory contains scripts for testing the Rust TPC-DS implementation against the Java reference implementation.
+This directory contains scripts for testing the Rust TPC-DS implementation
```
alamb, line 3: I reworked this readme to be less repetitive (removed all repetitive script descriptions in favor of documentation in the script headers)

tpcdsgen/scripts/README.md
```diff
@@ -1,285 +1,143 @@
...
 ./scripts/test-all-tables.sh
 ```
 
-## Scripts
-
-### 0. `bootstrap-java.sh` - Setup Java TPC-DS Implementation
+## Quick Start — C dsdgen conformance (`--compat c`)
 
-**⚠️ Run this first!** Sets up the Java TPC-DS implementation needed for conformance testing.
+The C reference data is pre-generated and published in
```
alamb, line 56: I spent quite a while messing around with this reference data with the original (very hard to work with) dsgen binary

tpcdsgen/scripts/README.md
```diff
@@ -1,285 +1,143 @@
...
-
-**Usage:**
-```bash
-# Generate all 25 tables (recommended first run)
-./scripts/generate-fixtures.sh
+### Tables excluded from automated checks
 
-# Generate specific tables
-./scripts/generate-fixtures.sh call_center warehouse
+The following tables are excluded from automated MD5 comparison; the
```
alamb, line 76: I am still messing around with one more table (that the datafusion-benchmarks repo manually converts a ISO-LATIN-1 file into utf8. 

I hope to regenerate this file without any modifications to the dsgen generator, and then we will avoid having to have this caveat

tpcdsgen/scripts/test-all-tables.sh
```diff
@@ -99,9 +120,38 @@ ALL_TABLES=(
     "web_site"
 )
 
-# Get list of tables to test
+# Tables to skip per compat mode (in addition to dbgen_version, which is
```
alamb, line 123: this is the same as above -- i am working to remove this

tpcdsgen/tests/fixtures/scale-1-c/MD5SUMS
```diff
@@ -1,25 +1,25 @@
 cc9aabc63eb8603bd7330b6735ed0961 call_center.dat
 0bbac1b8bdcf8ce2d5f0034980ee0196 catalog_page.dat
-8460b5abd6b6ceaf6107f217b016fb23 catalog_returns.dat
+f7cc7c90d74f59cbf73c36383eaf62cc catalog_returns.dat
```
alamb, line 3: rust MD5s aren't super useful as they are regenerated from source

this is the C MD5 sums

alamb:
FYI @clflushopt -- I pushed a bunch of changes to your branch, mostly related to the verification scripts 

What do you think about merging this PR into main and continuuing to iterate there?

clflushopt:
> FYI @clflushopt -- I pushed a bunch of changes to your branch, mostly related to the verification scripts
> 
> What do you think about merging this PR into main and continuuing to iterate there?

Works for me I'll do a last pass to review things and then merge into main. Thanks !

alamb:
> > FYI @clflushopt -- I pushed a bunch of changes to your branch, mostly related to the verification scripts
> > What do you think about merging this PR into main and continuuing to iterate there?
> 
> Works for me I'll do a last pass to review things and then merge into main. Thanks !

Thank you 🙏 

After that I will file some follow on tickets

There is one table I am still fighting (related to latin-1 encoding vs utf-8) but I will also do that as a follow on

alamb:
BTW I debugged the UTF8/LATIN 1 nonsense in customer
- https://github.com/apache/datafusion-benchmarks/issues/33

I am convinced that the data in tpcds-data is correct now  -- made from an  unmodified version of the original DSGen program

So once this PR is merged I will fix the code in this repo to match

tpcdsgen/src/config/scaling.rs
clflushopt: The Trino reference is [here](https://github.com/trinodb/tpcds/blob/8a02abbba864feedc2afd078c8153d66a95bb2d4/src/main/java/io/trino/tpcds/Table.java#L201)

clflushopt: Addressed

tpcdsgen/src/distribution/return_reasons_distribution.rs
clflushopt: The divergence is [here](https://github.com/trinodb/tpcds/blob/master/src/main/resources/io/trino/tpcds/distribution/return_reasons.dst#L38) I documented it in code as well.

clflushopt:
@alamb I addressed the comments for references let me know if there is anything else left to address otherwise feel free to resolve and merge.

tpcdsgen/src/distribution/return_reasons_distribution.rs
```diff
@@ -19,7 +19,9 @@ impl ReturnReasonsDistribution {
         })
     }
 
-    /// C dsdgen uses a corrected .dst where `reason 30` is not a duplicate of `reason 31`.
+    /// C dsdgen uses a corrected .dst where `reason 30` is not a duplicate of `reason 31`[1].
+    ///
+    /// [1]: https://github.com/trinodb/tpcds/blob/master/src/main/resources/io/trino/tpcds/distribution/return_reasons.dst#L38
```
alamb, line 24: Nice -- the trino port has a typo!

alamb (approved):
Love it

tpcdsgen/scripts/README.md
alamb: Update here is I think I have sorted the issue and have regenerated the data -- I think it is now correct (what is in datafusion-benchmarks is incorrect

- https://github.com/apache/datafusion-benchmarks/issues/33

alamb:
Follow up:

> 1. Pull out the CLI generation into its own crate (tpchdsgen-cli perhaps) similarly to tpchgen-cli

- Tracking in https://github.com/clflushopt/tpchgen-rs/issues/259

> 2. Create a tpchdsgen Arrow crate

- tracking in https://github.com/clflushopt/tpchgen-rs/issues/260

> 3. Rework the verification scripts so they don't have to use parquet (they can use pre-computed tpcds data files)

This was done in this initial PR

PR #254, opened 2026-05-10
chore: Follow-up improvements to parquet subcommand CLI

kevinjqliu:
Follow-up to #234. Updates docs, benchmarks, and CLI behavior for the `parquet` subcommand.

### CLI behavior
- Make `--format` optional (defaults to `tbl`) instead of requiring an explicit value
- Warn when `--format=parquet` is used, directing users to the `parquet` subcommand (v4.0.0 migration)
- Error when `--format` is used together with a subcommand
- Fix redundant "Warning:" prefix in `log::warn!` message

### Docs and benchmarks
- Update examples to use `parquet` subcommand instead of `--format=parquet`
- Add deprecation notice to README for `--format=parquet`, `--parquet-compression`, `--parquet-row-group-bytes`

### Tests
- Add integration tests for deprecation warning, format+subcommand conflict, and default format

### DIFF
--- a/benchmarks/BENCHMARKS.md
+++ b/benchmarks/BENCHMARKS.md
@@ -17,7 +17,7 @@ tpchgen-cli -s 100 --stdout | pv -arb > /dev/null
 # Outputs something similar to
 # 106GiB [3.09GiB/s] (3.09GiB/s)
 # For parquet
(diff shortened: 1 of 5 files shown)


clflushopt (approved):
lgtm thanks @kevinjqliu

PR #255, opened 2026-05-10
chore: Add tbl/csv subcommands, deprecate --format flag

### LINKED ISSUE #173: Investigate using clap subcommands to validate CLI arguments
I am curious to know if sub-commands in Clap allow options this way we can bind `--parquet` to a sub-command and these sub-cases can be handled that way instead. (Just a thought though, non-blocking and no need to address, this approach is great for now !)

_Originally posted by @clflushopt in https://github.com/clflushopt/tpchgen-rs/pull/158#discussion_r2280957410_

kevinjqliu:
Follow-up to #254. 
Closes #173.

Adds `tbl` and `csv` subcommands alongside the existing `parquet` subcommand so every output format is a first-class subcommand. The top-level `--format`, `--parquet-compression`, and `--parquet-row-group-bytes` flags are deprecated (removal in v4.0.0).

**Note that the default behavior remains unchanged**. Running without `--format` and without a subcommand (e.g. `tpchgen-cli -s 10`) still defaults to TBL output, preserving backward compatibility.

### New subcommands

- `tpchgen-cli tbl` — generates TBL (pipe-delimited) output
- `tpchgen-cli csv` — generates CSV output, with `--delimiter` option
- `tpchgen-cli parquet` — (already added in #254)

### Deprecated flags (hidden from `--help`, warn on use, removed in v4.0.0)

- `--format` — use `tbl`, `csv`, or `parquet` subcommand instead
- `--parquet-compression` — use `parquet --compression=...`
- `--parquet-row-group-bytes` — use `parquet --row-group-bytes=...`
-  `--delimiter` at top level — removed, move to csv subcommand; use `tpchgen-cli csv --delimiter=...` instead (we never released with `--delimiter`, so not a breaking change)

### Bug fix

- Mixing top-level args with a subcommand (e.g. `tpchgen-cli -s 10 parquet`) now errors instead of silently ignoring the top-level args (`args_conflicts_with_subcommands`)

- `csv --delimiter` rejects non-ASCII characters at parse time (e.g. `--delimiter='€'`) instead of failing mid-generation, since the underlying arrow-csv writer requires an ASCII byte

### DIFF
--- a/tpchgen-cli/bin/main.rs
+++ b/tpchgen-cli/bin/main.rs
@@ -344,7 +344,7 @@ impl Cli {
                 log::warn!("The --parquet-compression flag is deprecated. Use 'tpchgen-cli parquet --compression=...' instead");
                 builder = builder.with_parquet_compression(parquet_compression);
             } else {
-                log::warn!("Parquet compression option set but not generating Parquet files");
+                log::warn!("--parquet-compression ignored: output format is not parquet");
             }
         }
 
@@ -353,7 +353,7 @@ impl Cli {
                 log::warn!("The --parquet-row-group-bytes flag is deprecated. Use 'tpchgen-cli parquet --row-group-bytes=...' instead");
                 builder = builder.with_parquet_row_group_bytes(parquet_row_group_bytes);
             } else {
-                log::warn!("Parquet row group size option set but not generating Parquet files");
+                log::warn!("--parquet-row-group-bytes ignored: output format is not parquet");
             }
         }
 
--- a/tpchgen-cli/bin/main.rs
+++ b/tpchgen-cli/bin/main.rs
@@ -34,17 +34,13 @@ multiple files named <output_dir>/<table>/<table>.<part>.<format>
 
 Examples
 
-# Generate all tables at scale factor 1 (1GB) in TBL format to /tmp/tpch directory:
+# Generate all tables at scale factor 1 (1GB) in TBL format (default) to /tmp/tpch directory:
 
 tpchgen-cli -s 1 --output-dir=/tmp/tpch
 
-# Generate all tables in CSV format:
+# Generate all tables in CSV format with tab delimiter:
 
-tpchgen-cli -s 1 --format=csv --output-dir=/tmp/tpch
-
-# Generate scale factor one in CSV format with tab delimiter:
-
-tpchgen-cli -s 1 --format=csv --delimiter='\t' --output-dir=/tmp/tpch
+tpchgen-cli csv -s 1 --delimiter='\t' --output-dir=/tmp/tpch
 
 # Generate the lineitem table at scale factor 100 in 10 Apache Parquet files to
 # /tmp/tpch/lineitem:
@@ -66,6 +62,10 @@ struct Cli {
 
 #[derive(clap::Subcommand)]
 enum Commands {
+    /// Generate TBL (pipe-delimited) output
+    Tbl(TblArgs),
+    /// Generate CSV output with CSV-specific options
+    Csv(CsvArgs),
     /// Generate Apache Parquet output with Parquet-specific options
     Parquet(ParquetArgs),
 }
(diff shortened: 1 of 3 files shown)


clflushopt (approved):

kevinjqliu:
i thoroughly tested all the commands listed in this repo and all the error cases. 

all commands ran successfully locally
all error cases are covered by integration tests

tpchgen-cli/bin/main.rs
```diff
@@ -254,51 +300,47 @@ async fn main() -> io::Result<()> {
...
+                OutputFormat::Csv => {
+                    log::warn!("The --format flag will be removed in v4.0.0. Use `tpchgen-cli csv` instead.");
+                }
+                OutputFormat::Tbl => {
+                    log::warn!("The --format flag will be removed in v4.0.0. Use `tpchgen-cli tbl` instead.");
+                }
+            }
         }
 
+        let parquet_compression = self.args.parquet_compression.unwrap_or(Compression::SNAPPY);
```
clflushopt: question: Is defaulting to snappy going to be self-evident or should we pick "NoCompression" ? I ask because we already have a `if self.args.parquet_compression.is_some()` at line 332 ?

tpchgen-cli/bin/main.rs
```diff
@@ -254,51 +300,47 @@ async fn main() -> io::Result<()> {
...
                 log::warn!("The --parquet-compression flag is deprecated. Use 'tpchgen-cli parquet --compression=...' instead");
             } else {
                 log::warn!("Parquet compression option set but not generating Parquet files");
             }
         }
+
+        let parquet_row_group_bytes = self
+            .args
+            .parquet_row_group_bytes
+            .unwrap_or(DEFAULT_PARQUET_ROW_GROUP_BYTES);
```
clflushopt: question: same as my remark above we have a `self.args.parquet_row_group_bytes.is_some()` check at L:334

clflushopt (commented):
Left a couple on questions on how some args are `unwraped` prior to being checked via `is_some()` I think we should align that logic.

tpchgen-cli/bin/main.rs
kevinjqliu: good point, made the change not use `is_some`/ `unwraped`, instead falls back to `GeneratorConfig::default` if not specified

https://github.com/clflushopt/tpchgen-rs/blob/ee6c72ae597594db0bb412f59c67a17066dabaf0/tpchgen-cli/src/lib.rs#L241-L257

tpchgen-cli/bin/main.rs
kevinjqliu: same as above, fallback to `GeneratorConfig::default` if not specified

tpchgen-cli/bin/main.rs
```diff
@@ -181,7 +223,10 @@ struct ParquetArgs {
     row_group_bytes: i64,
 }
 
-/// Parse a delimiter string, handling escape sequences
+/// Parse a delimiter string, handling escape sequences.
+///
+/// The underlying arrow-csv writer requires an ASCII byte for the delimiter,
+/// so non-ASCII characters are rejected here rather than failing mid-generation.
```
kevinjqliu, line 229: found this bug, included to this pr for convenience

tpchgen-cli/bin/main.rs
```diff
@@ -112,27 +121,60 @@ struct CommonArgs {
...
+#[derive(clap::Args)]
+struct TblArgs {
+    #[command(flatten)]
+    common: CommonArgs,
+}
+
+#[derive(clap::Args)]
+struct CsvArgs {
+    #[command(flatten)]
+    common: CommonArgs,
```
kevinjqliu, line 177: note to reviewers: this diff is tricky. `delimiter` used to be top-level arg, now only part of csv arg

kevinjqliu:
@clflushopt im pretty happy with these changes, lmk if you have any other feedback!

kevinjqliu:
moving this forward, will fix forward if anything comes up 😄

PR #257, opened 2026-05-11
chore: Use pre-generated C data files, unify comparison scripts

alamb:
Note: This is still very much WIP, not ready for review

This targets 
- https://github.com/clflushopt/tpchgen-rs/pull/253

I totally got nerd-sniped today trying to verify the C compat mode 

Instead of downloading (slightly suspicious) parquet files, I think it would be better to use the raw, unencumbered output, from `dsgen` (the original C program)


alamb:
@clflushopt  I am taking the liberty of merging this consolidation directly into your branch so we can get it ready to go

PR #266, opened 2026-05-14
fix(tpcdsgen): Fix customer.dat to match C `dsdgen`

### LINKED ISSUE #261: [TPC-DS] Fix C inconsistency with "customer" table
As part of https://github.com/clflushopt/tpchgen-rs/pull/253 @clflushopt  implemented  the basic "C" compatibility mode

The rust code still makes one table that is still not quite the same as the C code, "customer" - and it has to do with non ascii character encoding

I have triple checked and I am now convinced that the data in the https://github.com/alamb/tpcds-data# repo is correct, and what was in datafusion-benchmarking was wrong. See 
- https://github.com/clflushopt/tpchgen-rs/pull/253#issuecomment-4443942364
- https://github.com/apache/datafusion-benchmarks/issues/33

Basically to fix this issue, we should remove this workaround, update the rust code, and get the test to pass:
https://github.com/clflushopt/tpchgen-rs/blob/5cde819660fa563b0f3517de40ffcf768ede412b/tpcdsgen/scripts/test-all-tables.sh#L123-L134

alamb:
## Summary

- Closes #261.

The Trino Java reference  writes `ISO_8859_1`, but the original C `dsdgen` writes in UTF-8

Adjust the output format so that the C compat mode output matches


tpcdsgen/src/output.rs
```diff
@@ -123,6 +121,41 @@ impl<W: Write> Write for Iso8859Writer<W> {
     }
 }
 
+/// Writer that selects the output encoding based on [`CompatMode`].
```
alamb, line 104: this is the key -- don't try and reencode the C data into ISO-8859

clflushopt (approved):
lgtm thanks !!

alamb:
Thanks gain @clflushopt

PR #286, opened 2026-06-28
refactor: Document and simplify tpcdsgen `Options` <--> `Session` conversion

alamb:
While reviewing the TPC-DS configuration code I found myself wishing that Options had more documentation.\n\nThis adds rustdoc for the CLI options and moves the Options/Session conversions onto standard Rust traits.\n\nIt also removes the parallel default constants by using clap's defaults directly.

### DIFF
--- a/tpcdsgen/src/config/options.rs
+++ b/tpcdsgen/src/config/options.rs
@@ -1,7 +1,13 @@
 use crate::config::{CompatMode, Session, Table};
-use crate::error::{InvalidOptionError, Result};
+use crate::error::{InvalidOptionError, Result, TpcdsError};
 use clap::Parser;
+use std::fmt;
 
+/// TPC-DS command-line options.
+///
+/// [`Options`] parses CLI arguments using [`clap`].
+/// Use [`Options::to_session`] to
+/// convert these values into a validated [`Session`] for row generation.
 #[derive(Parser, Debug, Clone)]
 #[command(name = "tpcdsgen")]
 #[command(about = "Rust implementation of TPC-DS data generator")]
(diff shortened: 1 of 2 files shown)


tpcdsgen/src/config/options.rs
```diff
@@ -56,69 +62,14 @@ pub struct Options {
 }
 
 impl Options {
-    // Default constants (matching Java implementation)
```
alamb, line 59: these are all redundant with the clap defaults, so remove them and use the clap values instead

tpcdsgen/src/config/options.rs
```diff
@@ -56,69 +62,14 @@ pub struct Options {
...
-            table: None,
-            null_string: Self::DEFAULT_NULL_STRING.to_string(),
-            separator: Self::DEFAULT_SEPARATOR.to_string(),
-            do_not_terminate: Self::DEFAULT_DO_NOT_TERMINATE,
-            no_sexism: Self::DEFAULT_NO_SEXISM,
-            parallelism: Self::DEFAULT_PARALLELISM,
-            overwrite: Self::DEFAULT_OVERWRITE,
-            compat: Self::DEFAULT_COMPAT,
-        }
+        Self::parse_from(["tpcdsgen"])
```
alamb, line 67: here is how to get the clap defaults

tpcdsgen/src/config/options.rs
```diff
@@ -56,69 +62,14 @@ pub struct Options {
...
-            table_option,
-            self.null_string.clone(),
-            separator_char,
-            self.do_not_terminate,
-            self.no_sexism,
-            self.parallelism,
-            self.overwrite,
-            self.compat,
-        ))
+        Session::try_from(self.clone())
```
alamb, line 72: moved to a more rust idomatic `TryFrom` impl

tpcdsgen/src/config/options.rs
```diff
@@ -184,6 +135,95 @@ impl Options {
...
+            options.do_not_terminate,
+            options.no_sexism,
+            options.parallelism,
+            options.overwrite,
+            options.compat,
+        ))
+    }
+}
+
+impl fmt::Display for Options {
```
alamb, line 178: Moved from     
```rust
pub fn get_command_line_arguments(&self) -> String {
```

As this logic is for converting Options to a command line argument.

And then I cleanly sepatated the logic that convertes from `Session` --> `Options` rather than combining the two

tpcdsgen/src/config/session.rs
```diff
@@ -193,49 +193,25 @@ impl Session {
...
-        if let Some(table) = self.table {
-            output.push(format!("--table {}", table.get_name()));
-        }
-        if self.null_string != Options::DEFAULT_NULL_STRING {
-            output.push(format!("--null {}", self.null_string));
-        }
-        if self.separator != Options::DEFAULT_SEPARATOR {
-            output.push(format!("--separator {}", self.separator));
+impl From<Session> for Options {
+    fn from(session: Session) -> Self {
```
alamb, line 201: instead of Session making the CLI arguments directly, now it first converts to a `Options` and then the options knows how to make CLI arguments, rather than mixing abstraction boundaries

tpcdsgen/src/config/options.rs
kevinjqliu: https://github.com/clflushopt/tpchgen-rs/blob/6d51156f32ffbb07b7eab3eca502556e8ea9092a/tpcdsgen/src/config/options.rs#L8-L56

tpcdsgen/src/config/options.rs
```diff
@@ -184,6 +135,95 @@ impl Options {
     }
 }
 
+impl TryFrom<Options> for Session {
+    type Error = TpcdsError;
+
+    fn try_from(options: Options) -> Result<Self> {
```
kevinjqliu, line 141: 👍  verbatim with `    pub fn to_session(&self) -> Result<Session> {` above

kevinjqliu (approved):
LGTM

alamb:
🚀  let's keep it moving

PR #276, opened 2026-06-21
feat: TPCDS <> Arrow support

clflushopt:
- closes https://github.com/clflushopt/tpchgen-rs/issues/260

PR #291, opened 2026-06-28
refactor: group TPC-DS table modules

kevinjqliu:
Follow up to #276, minor nit that can help with readability

Refactor the TPC-DS module layout so table-specific code lives together:

- Move `tpcdsgen-arrow` table Arrow generators under `src/tables/`
- Move `tpcdsgen` row structs and row generators under `src/row/tables/`
- Leave shared plumbing like `conversions.rs`, `row_generator.rs`, and `table_row.rs` at the existing shared-module level

This separates table-specific implementations from cross-table helpers, making the TPC-DS code easier to scan and future table changes easier to review. The refactor preserves the existing public re-exports and does not change generation behavior.

The code is easier to parse, now that we know which are helpers and which are the tables 😄 
`tpcdsgen-arrow/src`:
Screenshot 2026-06-28 at 10 52 15 AM

`tpcdsgen/src/row`:
Screenshot 2026-06-28 at 10 52 36 AM


alamb (approved):
looks good to me -- let's ship it!

PR #294, opened 2026-06-28
fix(tpcds): use `i32`/ `Date32` for  call_center columns

alamb:
- Closes #292

While reviewing the TPC-DS Arrow schemas, `call_center` was  inconsistent with the other slowly changing dimension tables.

This updates the Arrow representation for `call_center` so record date fields use `Date32` and date surrogate key fields use `Int64`, matching the pattern used by tables such as `store`, `web_site`, and `item`.

PR #295, opened 2026-06-29
fix(tpcds): use `i32` / `Date32` / `Time32` for dbgen_version create date 

alamb:
While working on #294, we also found that `dbgen_version.dv_create_date` and `dbgen_version.dv_create_time` were still stored as strings and exposed from `tpcdsgen-arrow` as `Utf8View`.

This keeps the generated `.dat` output unchanged, but carries `dv_create_date` as the typed TPC-DS `Date` and `dv_create_time` as seconds since midnight in the row model. The Arrow representation now exposes them as `Date32` and `Time32(Second)`.


kevinjqliu (approved):
LGTM 

verified against https://www.tpc.org/tpc_documents_current_versions/pdf/tpc-ds_v2.6.0.pdf

Image

alamb:
Thanks (again) @kevinjqliu

PR #300, opened 2026-06-29
chore: unify Arrow workspace dependency (on 59.0.0)

alamb:
The workspace currently uses two different Arrow versions: 58 in `tpcdsgen-arrow` and 59 elsewhere.

This adds workspace dependencies for Arrow 59 and updates the Arrow consumers to inherit them, keeping the Arrow crates on one version.


tpcdsgen-arrow/Cargo.toml
```diff
@@ -9,9 +9,9 @@ readme = "README.md"
 license = "Apache-2.0"
 
 [dependencies]
-arrow = { version = "58", default-features = false, features = ["prettyprint"] }
```
alamb, line 12: this was previously using 58

kevinjqliu (approved):
LGTM!

PR #290, opened 2026-06-28
Wire TPC-DS DAT generation into tpcgen

kevinjqliu:
## Summary
- Move the TPC-DS DAT generator out of `tpcdsgen/src/main.rs` and expose it as `tpcdsgen::dat`.
- Wire `tpcgen tpcds` and `tpcgen tpcds dat` to generate DAT output through the existing TPC-DS generator.
- Keep TPC-DS CSV and Parquet command forms explicitly not implemented.

## Notes for reviewers
- Commit 1 is a mechanical move into `tpcdsgen/src/dat.rs`.
- Commit 2 wires DAT generation into `tpcgen`.

## Validation
- `cargo test --offline -p tpcgen --test cli_integration -- --nocapture`
- `cargo build --offline --release -p tpcgen`
- `cargo clippy --offline -p tpcdsgen --lib --bin tpcdsgen -p tpcgen --bin tpcgen -- -D warnings`
- `git diff --check`

PR #303, opened 2026-06-29
[TPC-DS]: remove tpcdsgen CLI binary

alamb:
~Note: This isn't ready to review. since the conformance testing is tied up in the tpcds crate now, I think we also need to move the conformance testing scripts to the tpcgen crate. I'll try and do it tomorrow~

## Summary
- Remove the standalone `tpcdsgen` CLI binary so TPC-DS DAT generation goes through `tpcgen tpcds dat`.
- Remove external dependencies from the `tpcdsgen` crate by replacing Clap-derived config parsing and Chrono timestamp formatting with std-only code.
- Add `tpcgen tpcds dat --compat trino|c` and update conformance scripts/CI to call the unified CLI.

Follow-up to https://github.com/clflushopt/tpchgen-rs/pull/290#discussion_r3491122943.


.github/workflows/tpcdsgen-conformance.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
     - name: Build Rust table generators
       run: |
-        cargo build --release -p tpcdsgen
+        cargo build --release -p tpcgen
```
alamb, line 47: I will do the the same thing for the tpc-h tests as well

kevinjqliu:
looking forward to unify all the cli in `tpcgen`! will take another look when its ready for review

.github/workflows/tpcdsgen-conformance.yml
alamb: - see https://github.com/clflushopt/tpchgen-rs/issues/310

tpcdsgen/scripts/compare-table.sh
```diff
@@ -104,7 +104,7 @@ get_generator_for_table() {
     esac
 }
 
-# Find the unified tpcdsgen binary
+# Find the unified tpcgen binary
```
alamb, line 107: these scripts should really be in the tpcgen directory. This is tracked via
- https://github.com/clflushopt/tpchgen-rs/issues/310

tpcdsgen/src/config/compat.rs
```diff
@@ -1,9 +1,12 @@
+use std::fmt;
+use std::str::FromStr;
+
 /// Compatibility mode controlling which reference implementation to match.
 ///
 /// TPC-DS has two common reference implementations: the original C `dsdgen` and the
 /// Trino Java port. They diverge in a small number of places. This flag selects
 /// which behaviour to replicate.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, clap::ValueEnum)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
```
alamb, line 9: these structures were both used by the library as well as directly parsed by the CLI it seems (and thus depended on Clap)

This PR moves all the CLI logic (and clap dependency) into `tpcgen` and leaves the code in the `tpcdsgen` crate dependency free

tpcdsgen/src/config/options.rs
```diff
@@ -1,70 +1,78 @@
 use crate::config::{CompatMode, Session, Table};
 use crate::error::{InvalidOptionError, Result, TpcdsError};
-use clap::Parser;
 use std::fmt;
 
-/// TPC-DS command-line options.
+/// TPC-DS generation options.
```
alamb, line 5: Ditto here -- these are no longer used as part of a CLI

alamb:
This PR is now ready for review

kevinjqliu (approved):
Looks great, thanks!!

alamb:
The code 🚋  is moving right along

PR #313, opened 2026-06-30
[tpcgen]: move TPCH CLI tests into `tpcgen` (unified cli)

alamb:
- Follow-up to #296
- Part of https://github.com/clflushopt/tpchgen-rs/issues/218

We moved the code for tpchgen_cli -- now we should also move the tests 

Moves the TPCH CLI integration tests into the shared `tpcgen` crate and runs them as part of the existing `cli_integration` test binary.

Issue #218, opened 2026-01-10
[Tracking] TPC-DS support


clflushopt:
With #212 merged we currently have the machinery to fully support TPC-DS but it's not yet satisfying, this issue will track some changes and improvements we would like to make before claiming TPC-DS support.

Three items from @alamb recommendations are :

### Investigate the C implementation inconsistency

The port the Rust implementation was based on is [trino/tpcds](https://github.com/trinodb/tpcds) from their README

> C implementation. Therefore, when the C implementation has bugs (and it has many), we copy those bugs. Generally when copying over bugs, we include a comment for clarity.

> Testing
We unit test the tables by taking md5 hashes of the tables that we generate and comparing them to the md5sums of the tables generated by the tables generated by the dsdgen binary. We test the tables at all defined scales as well as at a scale that is in between the defined scales, to ensure that the row computation is correct for all scenarios.

The above suggests that the Java version should match the C version but from @alamb testing as reported [in this comment](https://github.com/clflushopt/tpchgen-rs/issues/51#issuecomment-3700913731) this is not the case. My hunch having dug around both codebases is that there is a bug we are not reproducing correctly but I am not sure its correct since our SF = 1 matches the Java version but not the C version. I will have to do a 3-way diff across all outputs to confirm.

### CLI driver integration

We will need to integrate the core generator logic with the `tpchgen-cli` this will probably be either easy or complex depending on what scaling model we will try and aim for. I also think the trick we introduced in #155 might not be trivial to re-use here due to how the scaling logic interacts with the stream synchronization TBC.

### Optimize the crap out of it
 
This will be our primary milestone we want to be as fast as the law of physics and computers allow us to be, I am already aware of a dozen low hanging fruits e.g. absorb all allocations out of the generation logic, reduce the amount of copying everywhere....

alamb:
I broke this ticket up into some subtasks to help organize the work a little more

Remaining subtasks 
- [x] https://github.com/clflushopt/tpchgen-rs/issues/261
- [x] https://github.com/clflushopt/tpchgen-rs/issues/259
- [x] https://github.com/clflushopt/tpchgen-rs/issues/260
- [ ] https://github.com/clflushopt/tpchgen-rs/issues/263

clflushopt:
Thanks @alamb

alamb:
BTW I have filed a bunch more tickets based on more recent work -- we are pretty closet to having this thing connected

[...]

PR #314, opened 2026-06-30
rename `tpcgen` crate and binary to `tpcgen-cli`

alamb:
- Follow-up to #313
- Part of https://github.com/clflushopt/tpchgen-rs/issues/218

the name `tpcgen` (which has a CLI) is confusing I think as it looks a lot like `tpchgen` and `tpcdsgen` which contain generator libraries

Screenshot 2026-06-30 at 7 58 53 AM

I propose renaming `tpcgen` to `tpcgen-cli` to 
1. Mirror the existing `tpchgen-cli` (aka we just drop the `h`)
2. Make it clear that `tpcgen-cli` contains CLI logic rather than generator logic

Renames the shared CLI crate from `tpcgen` to `tpcgen-cli` and renames the binary from `tpcgen` to `tpcgen-cli`.


tpcgen-cli/bin/tpcgen_cli.rs
```diff
@@ -1,13 +1,13 @@
 //! Binary entry point for the TPC-H and TPC-DS data generator.
 
 use clap::{Parser, Subcommand};
-use tpcgen::tpcds_cli::Cli as TpcdsCli;
-use tpcgen::tpch_cli::Cli as TpchCli;
+use tpcgen_cli::tpcds_cli::Cli as TpcdsCli;
```
alamb, line 4: now the `cli` suffix clearly marks what is CLI and what is not

alamb:
I am curious to know if this makes sense to you or not @kevinjqliu and @clflushopt

alamb:
One alternate is to just rename the crate (and not the binary) 🤔 

So the crate is called `tpcgen-cli` but the binary would still be `tpcgen` 🤔

kevinjqliu:
> One alternate is to just rename the crate (and not the binary) 🤔

I like the shorter form for the crate name (`tpcgen`) but I also value consistency between the crate and binary 🤔 

My final vote is for `tpcgen-cli` 😄 but could go either way

alamb:
> My final vote is for tpcgen-cli 😄 but could go either way

That is what this PR does :)

Let's wait to see what @clflushopt has to say

kevinjqliu:
oh no! i didnt mean to force push, just to fix the merge conflicts 😭

alamb:
> oh no! i didnt mean to force push, just to fix the merge conflicts 😭

no worries

kevinjqliu:
restored the original commit and resolved the merge conflicts on top

kevinjqliu (approved):
LGTM!

I checked both the original commit and the new commit, the shape of change looks the same

clflushopt (approved):
LGTM !

kevinjqliu:
moving this forward so i can start looking at some CLI improvements 😄 
Thanks again @alamb

alamb:
So exciting!@

PR #315, opened 2026-06-30
Run clippy on all targets and features

alamb:
I was running clippy locally and noticed a few failures which I wanted to fix.

This PR updates CI to run clippy across the full workspace with `--all-targets` and `--all-features`, so test code is checked as well.

It also fixes the new clippy errors in the `tpcdsgen` tests

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Run cargo-fmt
         run: cargo fmt --all -- --check
       - name: Run cargo-clippy
-        run: cargo clippy -- -D warnings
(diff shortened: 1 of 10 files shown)


.github/workflows/rust.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Run cargo-fmt
         run: cargo fmt --all -- --check
       - name: Run cargo-clippy
-        run: cargo clippy -- -D warnings
+        run: cargo clippy --workspace --all-targets --all-features -- -D warnings
```
alamb, line 20: this now runs clippy on the test code as well as the library code

kevinjqliu (approved):
LGTM!

Issue #309, opened 2026-06-30
[TPC-DS] Support generating TPC-DS data as CSV


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

This should work and generate the TPC-DS data in CSV form:

```shell
andrewlamb@Andrews-MacBook-Pro-3:~/Software/tpchgen-rs$ ./target/debug/tpcgen tpcds csv
Error: TPC-DS data generation is not yet implemented
```

PR #317, opened 2026-07-01
refactor: Move some output configuration and writing from tpcdsgen

alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/309

Right now the output configuration (what file to write to, what format, etc) is
intermixed as part of the "core tpcds generator" crate. This is due to the
original design that had a CLI in the same crate

To cleanly support multiple output formats CSV and parquet I want to separate
the core generator logic from the logic that that writes the output.

This PR moves a bunch of the output configuration into the cli and out of tpcdsgen (note it is a net removal of code)


tpcdsgen/src/config/options.rs
```diff
@@ -1,396 +0,0 @@
-use crate::config::{CompatMode, Session, Table};
-use crate::error::{InvalidOptionError, Result, TpcdsError};
-use std::fmt;
-
-/// TPC-DS generation options.
```
alamb, line 5: The "options" struct was basically a CLI parsing struct. Now that we have a unified CLI in tpcgen, we can remove this structure in favor of the CLI in tpcgen

tpcdsgen/src/config/session.rs
```diff
@@ -22,11 +22,35 @@ pub struct Session {
 
 impl Default for Session {
     fn default() -> Self {
-        Session::get_default_session()
+        Self::new_with_chunk_number(
```
alamb, line 25: I plan to clean up Session in a follow on PR. I think if we had a Builder style API to create sessions, this code would be a lot simpler.

tpcdsgen/src/config/session.rs
```diff
@@ -61,12 +85,84 @@ impl Session {
         )
     }
 
+    /// Create and validate a session for a single generation chunk.
```
alamb, line 88: I will simplify this with a Builder style API as a follow on PR, but for now I am trying to keep the PRs smaller and  easy to review

tpcdsgen/src/config/session.rs
```diff
@@ -247,25 +338,49 @@ impl Session {
     ///
     /// Default-valued options are omitted from the returned string.
     pub fn get_command_line_arguments(&self) -> String {
```
alamb, line 340: I will also simplify this by simply passing in the CLI arguments -- as a follow on PR

tpcgen/src/tpcds_cli.rs
```diff
@@ -171,18 +173,33 @@ impl CommonArgs {
...
+        let table = table
+            .as_deref()
+            .map(|table| {
+                table
+                    .parse::<Table>()
+                    .map_err(|_| InvalidOptionError::new("table", table))
+            })
+            .transpose()?;
+
+        Ok(Session::try_new(
```
alamb, line 190: now we make a Session directly rather than Options --> Session

tpcdsgen/src/config/session.rs
alamb: - see https://github.com/clflushopt/tpchgen-rs/pull/318

kevinjqliu (approved):
LGTM!

tpcgen/src/tpcds_cli.rs
clflushopt: this will be very useful when we parallelize generators thnks !

clflushopt (approved):

alamb:
Thanks @kevinjqliu and @clflushopt

PR #318, opened 2026-07-01
[tpcdsgen]: add validated SessionBuilder

alamb:
- stacked on https://github.com/clflushopt/tpchgen-rs/pull/317
- part of https://github.com/clflushopt/tpchgen-rs/issues/218
- part of https://github.com/clflushopt/tpchgen-rs/issues/309

As part of separating the CLI / output configuration from the core TPC-DS row generation logic, `Session` creation is  pretty messy. 

This PR adds a validated `SessionBuilder` API for constructing `Session` values and removes a bunch of `Session` constructors and convenience modifiers. 

This PR also removes regenerated command line arguments from `Session` and instead stores the actual command line arguments supplied by the `tpcgen` CLI when available. The `dbgen_version` table now uses that stored value rather than reconstructing arguments from session fields.

PR #319, opened 2026-07-01
[tpcdsgen]: move output options out of Session

alamb:
- stacked on https://github.com/clflushopt/tpchgen-rs/pull/318
- part of https://github.com/clflushopt/tpchgen-rs/issues/309

As part of separating CLI / output configuration from the core TPC-DS row generation logic, `Session` should only carry settings needed to build rows. Output-specific configuration belongs in the CLI layer.

This PR removes output fields from `tpcdsgen::config::Session`, including target directory, file suffix, null string, separator, row termination, parallelism, and overwrite configuration.

The TPC-DS DAT CLI now owns those settings through a DAT output options struct and passes them to the DAT writer alongside the core `Session`.


tpcdsgen/src/config/session.rs
```diff
@@ -25,16 +18,9 @@ impl Default for Session {
     fn default() -> Self {
         Session {
             scaling: Scaling::new_with_compat(Self::DEFAULT_SCALE, Self::DEFAULT_COMPAT),
-            target_directory: Self::DEFAULT_DIRECTORY.to_string(),
```
alamb, line 28: these fields are for output, not generation, so move them to the CLI

tpcgen-cli/src/tpcds_cli/dat.rs
```diff
@@ -18,30 +18,100 @@
...
 use tpcdsgen::config::{Session, Table};
+use tpcdsgen::error::InvalidOptionError;
 use tpcdsgen::output::CompatWriter;
-use tpcdsgen::row::*;
+use tpcdsgen::row::{GeneratedRow, TableRow, *};
 use tpcdsgen::types::Date;
 
 type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
+/// DAT output configuration owned by the CLI layer.
```
alamb, line 32: Now all output options live in the CLI code (where they are used to write .dat files)

When we support CSV output, some of this code will likely need to be refactored again (as most if it can be shared with CSV output) but for now I thought it was best to leave it in its own file

tpcgen-cli/src/tpcds_cli/dat.rs
```diff
@@ -360,10 +460,44 @@ fn generate_inventory(session: &Session) -> Result<()> {
...
 }
+
+fn create_output_file(path: &Path, output_options: &OutputOptions) -> Result<File> {
+    if !output_options.overwrite && path.exists() {
+        return Err(format!("Output file {} already exists", path.display()).into());
+    }
+    Ok(File::create(path)?)
+}
+
+fn write_row(
```
alamb, line 478: I am salivating at how much faster we can make this, but I am holding off

alamb: Looks like @clflushopt already did it in 
- https://github.com/clflushopt/tpchgen-rs/pull/320

tpcdsgen-arrow/tests/reparse.rs
```diff
@@ -57,7 +58,7 @@ fn test_row_count(table: Table) -> i64 {
 fn parse_dat<'a>(data: &'a [u8], schema: &'a SchemaRef) -> impl Iterator<Item = RecordBatch> + 'a {
     let null_re = regex::Regex::new("^$").unwrap();
     let builder = arrow_csv::reader::ReaderBuilder::new(Arc::clone(schema))
-        .with_delimiter(SESSION.get_separator() as u8)
+        .with_delimiter(DAT_SEPARATOR as u8)
```
alamb, line 61: the point of this PR is to move the output specific stuff (controlled by the CLI) out of the dsgen crate 

In this case, the choice of separator is not part of the core dsgen crate

alamb:
FYI @kevinjqliu  and @clflushopt  -- this is my last refactoring PR to move the output configuration into the CLI code and out of the core dsgen crate

I am pretty sure this will conflict with 
- https://github.com/clflushopt/tpchgen-rs/pull/320

but I am happy to resolve the conflicts

tpcgen-cli/src/tpcds_cli/dat.rs
alamb: I think it eventually should likely be part of the CLI options if we want to expose them (and there are some we probably will just remove like `parallelism` and `suffix`

tpcgen-cli/src/tpcds_cli/dat.rs
```diff
@@ -360,10 +460,44 @@ fn generate_inventory(session: &Session) -> Result<()> {
...
+        output_options.suffix
     ))
 }
+
+fn create_output_file(path: &Path, output_options: &OutputOptions) -> Result<File> {
+    if !output_options.overwrite && path.exists() {
+        return Err(format!("Output file {} already exists", path.display()).into());
+    }
+    Ok(File::create(path)?)
+}
```
kevinjqliu, line 476: yesss!! this one has been bothering me 😄

kevinjqliu (commented):
Looks like some of the `OutputOptions` are not configurable through the CLI anymore. Is that intentional? 

Only `target_directory` is configurable, via `--output-dir`
https://github.com/clflushopt/tpchgen-rs/pull/319/changes#diff-6bda1a13812f0482db3f339ee700bd449feb9b54114967ec237ce3998a130db8R164

These fields are no longer configurable from the binary:
- `suffix`: default to `.dat`
- `null_string`: default to `""`
- `separator`: default to `|`
- `do_not_terminate`: default to `false`
- `parallelism`: default to `1`
- `overwrite`: default to `true`

alamb:
> These fields are no longer configurable from the binary:


I don't think they are currently configurable (via the tpcgen-cli) either:

https://github.com/clflushopt/tpchgen-rs/blob/ea511673dbec41b7adc6832c5d44917b6266faf4/tpcgen-cli/src/tpcds_cli.rs#L36-L38

https://github.com/clflushopt/tpchgen-rs/blob/ea511673dbec41b7adc6832c5d44917b6266faf4/tpcgen-cli/src/tpcds_cli.rs#L94


I am thinking that we will follow the same set of parameters at tpch unless someone really wants more control. 

I think the main usecase for this crate will be creating .dat files (whatever the defautls of dsgen were) and parquet -- I am doubtful if people really care about changing the suffix or null string, for example

I propose we simply don't support these ones:

```
suffix: default to .dat
null_string: default to ""
separator: default to |
do_not_terminate: default to false
```

I propose we replace these with flags similar to tpchgen
```
parallelism: default to 1
overwrite: default to true
```

kevinjqliu:
sounds good! we can stick with this for now and evolve if there are future use cases

alamb:
(any concerns about merging this @kevinjqliu ?)

kevinjqliu:
🚢  it !

alamb:
woohoo!

PR #320, opened 2026-07-01
feat: TPCDS optimizations (single core)

clflushopt:

This PR introduces 2 changes to speed-up TPCDS core generator logic by optimizing out some allocations from the hot-path. Profiles captured are attached, I split the changes to be per-commit so they are easy to reproduce and profile.

- Changes: 
  - Drop `core::fmt` machinery when calling `Iso8859Writer::write` (39%)
  - Eliminate the C-style float logic via `f64` in `Decimal::Display` (16%)
  - Introduce `Line` buffer wrapper around `Vec<u8>` to avoid allocations from `Vec<String>` to represent a datum row. (58%)
  - Cache scaling counts table (15%)

In total SF1 wall clock time is down from 17.3s to 3.1s ~ 5.5x speed up overall. Most of these changes are just low hanging fruits, I believe there is still much to improve single-core perf wise before we even parallelize generation which with #319 would become easier.

- Profile for `main`: 
tpcds-flamegraph-c2047c2

- `Iso8859Writer::write` fast path :
tpcds-flamegraph-3a03494

- `Decimal::Display` change: 
tpcds-flamegraph-8fd5362


kevinjqliu (approved):
nice!

clflushopt:
With [27e74a8fd399ad424a216fdb8b5b2da010039cda](https://github.com/clflushopt/tpchgen-rs/pull/320/changes/27e74a8fd399ad424a216fdb8b5b2da010039cda) now SF1 generation drops by about 2x from 8.9s to 4.7s I also added a profiling script which generates flamegraph and a perf-compatible report that can be viewed in https://profiler.firefox.com/

[firefox.perf.gz](https://github.com/user-attachments/files/29610507/firefox.perf.gz)
flamegraph

benchmarks/profile.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
+#
+# Build and capture a commit-tagged CPU profile of data generation.
```
alamb, line 3: FWIW I have had really good luck using `samply` https://github.com/mstange/samply

```shell
cargo install --locked samply
```

It can make flamegraphs as well as many others

so like 
```shell
samply record ./target/release/tpcgen-cli tpcds
```

Image

Image

You might be able to use it instead of a custom set of scripts

tpcdsgen/src/config/scaling.rs
```diff
@@ -25,7 +35,20 @@ impl Scaling {
     }
 
     pub fn new_with_compat(scale: f64, compat_mode: CompatMode) -> Self {
-        Scaling { scale, compat_mode }
+        let mut scaling = Scaling {
```
alamb, line 38: maybe worth making this `Scaling::new()`

tpcdsgen/src/config/scaling.rs
```diff
@@ -25,7 +35,20 @@ impl Scaling {
...
+            scale,
+            compat_mode,
+            row_counts: [0; CACHED_TABLE_COUNT],
+        };
+        for table in Table::main_tables() {
+            // Inventory derives from the Item and Warehouse entries, so it
+            // is filled once the rest of the cache is populated.
+            if table != Table::Inventory {
+                scaling.row_counts[table as usize] = scaling.compute_row_count(table);
```
alamb, line 47: If you wanted to encapsulate this more, you could add a method to `Table` like `table.index()` or something that captures the `usize`

tpcdsgen/src/config/scaling.rs
```diff
@@ -38,12 +61,23 @@ impl Scaling {
...
     /// Get row count for a table at this scale factor.
     ///
-    /// Uses the table's ScalingInfo to properly calculate row counts based on
-    /// the scaling model (Static, Linear, or Logarithmic).
+    /// Main output tables are answered from the cache built at construction;
+    /// only source tables fall through to the model computation (where they
+    /// panic, as before).
+    pub fn get_row_count(&self, table: Table) -> i64 {
+        let index = table as usize;
+        if index < CACHED_TABLE_COUNT {
```
alamb, line 69: I wonder if it would make sense to just pre-compute the row counts for all tables up front (even though the non fact tables aren't performance critical) to keep the code more uniform (so row count is always just a look up)

tpcdsgen/src/config/scaling.rs
```diff
@@ -38,12 +61,23 @@ impl Scaling {
...
+        self.compute_row_count(table)
+    }
+
+    /// Compute a table's row count from its ScalingInfo model (Static,
+    /// Linear, or Logarithmic). Used to populate the cache.
     ///
     /// Note: Inventory is a special case - its row count is computed dynamically
     /// as item_id_count × warehouse_count × weeks (matching Java's scaleInventory()).
-    pub fn get_row_count(&self, table: Table) -> i64 {
+    fn compute_row_count(&self, table: Table) -> i64 {
```
alamb, line 80: It might be nice to add a test too that verifies the row counts reported by Scaling match the row counts computed directly

tpcdsgen/src/row/tables/catalog_returns_row.rs
```diff
@@ -120,6 +121,58 @@ impl CatalogReturnsRow {
         }
     }
 
+    fn push_key(
```
alamb, line 124: these methods all seem very similar, and they only use `self` to check for null.

We might be able to reduce the replication by putting the methods on `Line` directly -- for example

```rust
impl Line {
  fn push_key(
    &mut self, 
    is_null: bool,
    sep: &[u8],
    value: i64,
  )
...
}
```

tpcdsgen/src/output.rs
```diff
@@ -88,11 +199,38 @@ impl<W: Write> Iso8859Writer<W> {
 /// Each UTF-8 character is converted to its ISO-8859-1 equivalent.
 impl<W: Write> Write for Iso8859Writer<W> {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        // Interpret input as UTF-8, convert to ISO-8859-1
+        // ASCII bytes are identical in UTF-8 and ISO-8859-1, so they can be
+        // passed through untouched. TPC-DS output is almost entirely ASCII,
+        // making this the hot path.
```
alamb, line 204: thiis is a good idea

tpcdsgen/src/output.rs
```diff
@@ -88,11 +199,38 @@ impl<W: Write> Iso8859Writer<W> {
...
+            self.inner.write_all(buf)?;
+            return Ok(buf.len());
+        }
+        // Interpret input as UTF-8, convert to ISO-8859-1 through a stack
+        // buffer to avoid heap allocations on every write.
         let s =
             std::str::from_utf8(buf).map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
-        let iso_bytes = to_iso_8859_1(s)?;
-        self.inner.write_all(&iso_bytes)?;
+        let mut chunk = [0u8; 256];
```
alamb, line 213: using a bgger buffer here might also help (e.g. 1024 or 4096 bytes rather than 256) to amortize the write call more

tpcdsgen/src/output.rs
```diff
@@ -50,6 +50,118 @@ pub fn to_iso_8859_1(s: &str) -> io::Result<Vec<u8>> {
         .collect()
 }
 
+/// A reusable line buffer for the hot row serialization path
```
alamb, line 53: This is a definite improvement over the current situation. However, I think it still requires an extra copy (a write to line and then a write to the actual output)

Longer term I suggest we move towards using `std::fmt::Display` so we can write directly into an output buffer. 

The pattern which is used in tpchgen is to implement std::fmt::Display, and then use a BufWriter to buffer into memory when we want to avoid a syscall for each write call

tpcdsgen/src/row/tables/catalog_returns_row.rs
alamb: Actually, after reading more of this PR, I actually suggest following the same model as we did for tpch

For example

https://github.com/clflushopt/tpchgen-rs/blob/f7e49ff4533d9ae432f3819354ee7564f5f406cc/tpchgen/src/generators.rs#L1179-L1187

With each row type knows how to write itself to a std::fmt:Formatter, hard coding the `|` separator to start.

Then in our CLI we have special wrapper over each row that knows how to write as CSV or TBL, for example, again directly into the output via std::fmt

https://github.com/clflushopt/tpchgen-rs/blob/9fdba0435ac338179697cf06d0da00b42b2a9827/tpchgen/src/csv.rs#L30-L69

tpcdsgen/src/types/decimal.rs
```diff
@@ -144,26 +144,55 @@ impl Decimal {
     pub fn get_number(&self) -> i64 {
         self.number
     }
+
+    /// Appends the same representation as `Display` to `out` without going
```
alamb, line 148: This is largely duplicated with the Display implementation below -- I think we could avoid the duplication by standardizing on using std::fmt::Display everywhere (though that is a larger refactor of this PR)

alamb (approved):
Thank you @clflushopt -- this looks like a great step forward. 5x faster 😎 

I left a bunch of suggestions and about how to potentially rework this to be quite a bit less code and perhaps even faster, but I think this is a lot faster so we can/should go with it for now. 

# My Benchmarking  (confirms 5x faster!)

Command I ran (I love `hyperfine`, thank you @kevinjqliu  for introducing me to it)

```shell
  cargo build --release -p tpcgen --bin tpcgen

  hyperfine --warmup 1 --runs 3 \
    --prepare 'rm -rf /tmp/tpcdsgen-sf1' \
    './target/release/tpcgen tpcds dat --scale-factor 1 --output-dir /tmp/tpcdsgen-sf1 --no-progress'
```

###  Main (`git merge-base HEAD origin/main`)

Benchmark 1: ./target/release/tpcgen tpcds dat --scale-factor 1 --output-dir /tmp/tpcdsgen-sf1 --no-progress
  Time (mean ± σ):     24.559 s ±  0.207 s    [User: 23.997 s, System: 0.484 s]
  Range (min … max):   24.429 s … 24.798 s    3 runs


### This branch
```
Benchmark 1: ./target/release/tpcgen tpcds dat --scale-factor 1 --output-dir /tmp/tpcdsgen-sf1 --no-progress
  Time (mean ± σ):      4.713 s ±  0.020 s    [User: 4.101 s, System: 0.601 s]
  Range (min … max):    4.691 s …  4.730 s    3 runs
```

alamb:
I also merged this PR up to resolve a conflict

clflushopt:
> I also merged this PR up to resolve a conflict

I fixed the merge conflict and I am going trough the comments to address them and clean it up a bit ! Thanks for the review

alamb (commented):
I feel bad about dropping a large amount of feedback on this PR and I fear it is going to start gathering (even more) merge conflicts 

@clflushopt what do you think about breaking this PR into smaller  PRs? 

For example the `Scaling` cache would probably be a good / easy to extract smaller PR.

Then maybe  we could focus on one table (e.g. `store_sales`, the biggest), and figure out a good pattern for optimized CSV and DAT generation.

clflushopt:
@alamb agreed I also decided to just do the refactor to `fmt::Display` approach directly instead of doing the refactor here.

clflushopt:
Closed in favor of #359 and its followups

Issue #287, opened 2026-06-28
fix pypi publishing


kevinjqliu:
> Not sure how we will test this other than "in production" after the next release, but that is ok with me! 

 _Originally posted by @alamb in [#280](https://github.com/clflushopt/tpchgen-rs/pull/280/changes#r3487985808)_

Publishing on `release` trigger seems unsafe. We can make it just manually triggered. 

Thats also a clever trick to override the python package version, https://github.com/apache/iceberg-python/blob/9d4cefb6adefe4894490288a913b292562df1539/.github/workflows/pypi-build-artifacts.yml#L22-L27

So we can test using pre-release version tags

kevinjqliu:
same issue in `.github/workflows/tpchgen-cli-publish-pypi.yml`, they mirror each other

kevinjqliu:
for [`tpchgen-publish-crates.yml`](https://github.com/clflushopt/tpchgen-rs/blob/main/.github/workflows/tpchgen-publish-crates.yml)
* add new tpcds* crates
* make sure publish ordering is correct
* maybe rename  to just `publish-crates`

kevinjqliu:
https://github.com/clflushopt/tpchgen-rs/pull/326 fixes the crates.io side
- manually published to crates.io for `tpcdsgen` / `tpcdsgen-arrow` / `tpcgen-cli` as pre-release version so that the github workflow can publish using trusted publisher
- edit all crates on crates.io to use the new `publish-crates.yml` file in trusted publisher setting

pypi side is easier, we can configure a "Pending publishers" and the `tpcgen-cli-publish-pypi.yml` workflow will work as is. I've already configured it on PyPi:
<img width="806" height="178" alt="Image" src="https://github.com/user-attachments/assets/2ea11967-87f6-44f3-92bb-2ca9cedf2ca4" />

Last optional todo is to be able to trigger the crates.io/pypi workflow as pre-release version

kevinjqliu:
> Last optional todo is to be able to trigger the crates.io/pypi workflow as pre-release version

we dont necessary need this, can just manually publish (very carefully)

PR #322, opened 2026-07-03
Fix release workflow triggers and permissions

kevinjqliu:
Part of #287

## Why

`release.created` can miss draft releases when they are later published. `release.published` matches the actual publish moment.

When a job declares a custom permissions block, GitHub disables every permission that is not listed. These jobs still need repository read access for checkout, so this change lists `contents: read` explicitly.


alamb (approved):
seems plausible to me (and claude and codex). Let's give it a go!

alamb:
merged up to resolve a conflict

Issue #302, opened 2026-06-29
[TPC-DS] integrate progress tracker to CLI


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

@kevinjqliu is integrating the progress tracker locally.

> This looks beautiful 😄 

https://github.com/user-attachments/assets/737c01a1-f50a-4591-bacd-2776ca6063e9

_Originally posted by @kevinjqliu in https://github.com/clflushopt/tpchgen-rs/issues/290#issuecomment-4827276187_

PR #323, opened 2026-07-03
Make progress tracker API always available

### LINKED ISSUE #298: Rework `progress` and `indicatif-progress` feature flags
After https://github.com/clflushopt/tpchgen-rs/pull/226 from @kevinjqliu  the crate has a very neat progress bar!

However, the crate now also has 2 features: `progress` and `indicatif-progress`

```toml
[features]
default = ["indicatif-progress"]
# Enable progress callbacks for applications that want to provide their own
# progress reporting without pulling in the default terminal progress bars.
progress = []
# Enable the default CLI terminal progress bars backed by `indicatif`.
indicatif-progress = ["progress", "dep:indicatif"]
```

Which results in an unfortunate number of feature gates like this
```rust
#[cfg(feature = "progress")]
```
and 
```rust
#[cfg(not(feature = "progress"))]
```

I think the main rationale was to keep dependencies down (and keep indicatif a optional dependency). 

## Proposal
1. Remove the "progress" feature and always include the progress reporting API / traits
2. When `indicatif-progress` is NOT enabled then there will be a stub progress reporting implementation that doesn't do anything
3. When `indicatif-progress` is NOT enabled then there will be a progress reporting implementation that uses `indicatif` to report progress

kevinjqliu:
- Closes #298

- Remove the `progress` feature and always include the progress reporting API / traits
- When `indicatif-progress` is NOT enabled then there will be a stub progress reporting implementation that doesn't do anything
- When `indicatif-progress` is enabled then there will be a progress reporting implementation that uses `indicatif` to report progress

Overall improves the code quality of Progress tracker implementation (https://github.com/clflushopt/tpchgen-rs/commit/5f652feebc46fa788a8a2f0ea76813f69de74797)
Note that removing `progress` is a breaking change (since it was released as part of v3.0.0), but i think its fine to do so

### DIFF
--- a/tpcgen-cli/src/tpch_cli/generate.rs
+++ b/tpcgen-cli/src/tpch_cli/generate.rs
@@ -199,7 +199,7 @@ impl BufferRecycler {
     }
 }
 
-#[cfg(all(test, feature = "progress"))]
+#[cfg(test)]
 mod tests {
     use super::*;
     use crate::tpch_cli::progress::{ProgressTracker, RunProgress};
--- a/tpcgen-cli/src/tpch_cli/generate.rs
+++ b/tpcgen-cli/src/tpch_cli/generate.rs
@@ -59,25 +59,10 @@ where
     I: Iterator<Item = G>,
     S: Sink + 'static,
 {
-    let progress = Default::default();
-    generate_in_chunks_impl(sink, sources, num_threads, progress).await
+    generate_in_chunks_with_progress(sink, sources, num_threads, TableProgress::default()).await
 }
 
 pub(crate) async fn generate_in_chunks_with_progress<G, I, S>(
-    sink: S,
-    sources: I,
-    num_threads: usize,
-    progress: TableProgress,
-) -> Result<(), io::Error>
-where
-    G: Source + 'static,
-    I: Iterator<Item = G>,
-    S: Sink + 'static,
-{
-    generate_in_chunks_impl(sink, sources, num_threads, progress).await
-}
-
-async fn generate_in_chunks_impl<G, I, S>(
     mut sink: S,
     sources: I,
     num_threads: usize,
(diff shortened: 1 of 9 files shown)


kevinjqliu:
net code removal and it look cleaner, thanks for the suggestion @alamb

kevinjqliu:
#302 should also be easier after this PR 😄

tpcgen-cli/src/tpch_cli/generate.rs
kevinjqliu, line 52: simplified these helpers from 3 -> 2:
- `generate_in_chunks` remains the public API.
- `generate_in_chunks_with_progress` is the crate-private API used by `runner.rs`.

This keeps progress plumbing internal while preserving the existing public call shape.

tpcgen-cli/src/tpch_cli/generator.rs
```diff
@@ -226,7 +224,6 @@ impl Default for GeneratorConfig {
 /// Use the builder pattern via [`TpchGenerator::builder()`] to configure and create instances.
 pub struct TpchGenerator {
     config: GeneratorConfig,
-    #[cfg(feature = "progress")]
```
alamb, line 229: this looks much cleaner to me

tpcgen-cli/src/tpch_cli/generator.rs
```diff
@@ -226,7 +224,6 @@ impl Default for GeneratorConfig {
...
     progress_tracker: Option<Arc<dyn ProgressTracker>>,
```
alamb, line 227: We can potentially, as a follow on PR, make this even simpler by changing to always having a progress_tracker, and by default installing a `NoOpProgressTacker` or something -- that way the code doesn't have to check the Option.

```rust
    progress_tracker: Arc<dyn ProgressTracker>,
```

alamb (approved):
Thanks @kevinjqliu  -- this is a great simplification in my mind!

I also tested locally with

```shell
cargo run --bin tpcgen-cli --release -- tpch parquet --scale-factor=10
```
```shell
argo run --no-default-features  --bin tpcgen-cli --release -- tpch parquet --scale-factor=10
```

And it still had progress/no progress as expected

tpcgen-cli/src/tpch_cli/generator.rs
kevinjqliu: great suggestion! created https://github.com/clflushopt/tpchgen-rs/pull/325

PR #340, opened 2026-07-06
[tpcgen-cli] Enable TPC-DS parquet logging

alamb:
## Summary

Enable the TPC-DS Parquet CLI path to initialize logging, matching the DAT and CSV paths so `--verbose` and `--quiet` are applied consistently.

This fixes a parity gap where `tpcgen-cli tpcds parquet -v` accepted the flag but never called `configure_logging`, so it didn't do anything


kevinjqliu (approved):
LGTM!

PR #341, opened 2026-07-06
[tpcgen-cli] Respect TPC-DS parquet row group bytes

### LINKED ISSUE #338: [TPC-DS] Parquet output should respect the parquet row group size

 `tpcgen tpcds parquet --row-group-bytes ...` is accepted, but the TPC-DS Parquet writer only sets  compression.

We should respect the command line argument

alamb:
## Summary
- closes  #338

Respect `tpcgen-cli tpcds parquet --row-group-bytes` 

Before this change the CLI accepted `--row-group-bytes` for TPC-DS Parquet output, but discarded it before constructing the writer, so all files used the writer default row group layout.

The regression test follows the same pattern as the existing TPC-H row group tests: https://github.com/clflushopt/tpchgen-rs/blob/main/tpcgen-cli/tests/cli_integration/tpch.rs#L443-L568


tpcgen-cli/tests/cli_integration/tpch.rs
```diff
@@ -780,44 +780,6 @@ fn read_reference_file(table_name: &str, scale_factor: &str) -> String {
     }
 }
 
-#[derive(Debug, PartialEq)]
```
alamb, line 783: Refactored into a common module so it can be shared by tests

tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -158,8 +158,22 @@ impl CsvArgs {
 
 impl ParquetArgs {
     fn run(self) -> Result<()> {
-        let _ = self.row_group_bytes;
-        self.common.run_parquet(self.compression)
+        if self.row_group_bytes <= 0 {
```
kevinjqliu, line 161: nit: would be nice for ParquetArgs / clap to handle the arg validation

kevinjqliu (approved):
LGTM!

PR #344, opened 2026-07-06
[tpcgen-cli] Remove dead code in TPCH parquet generation

alamb:
## Summary

Remove the no-progress `generate_parquet` wrapper and rename the progress-aware TPCH parquet helper to `generate_parquet`.

All TPCH parquet call sites now pass an explicit progress tracker and table name, so parquet generation follows the same explicit progress contract as the rest of the TPCH writer path.

### DIFF
--- a/tpcgen-cli/src/tpch_cli/parquet.rs
+++ b/tpcgen-cli/src/tpch_cli/parquet.rs
@@ -1,6 +1,6 @@
 //! Parquet output format
 
-use crate::tpch_cli::progress::{no_op_progress_tracker, ProgressTracker};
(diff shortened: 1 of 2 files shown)


tpcgen-cli/src/tpch_cli/parquet.rs
alamb, line 31: nothing calls this function -- I think it was to keep the API the same

kevinjqliu (approved):
LGTM! 

I think it was originally implemented that way to keep the original function signature backwards compatible. But since we're making breaking changes for the upcoming release, this is much much cleaner. 
Thanks!!

PR #342, opened 2026-07-06
[tpcgen-cli]: extract shared parquet writer from tpch generation

alamb:
## Why
- part of https://github.com/clflushopt/tpchgen-rs/issues/334

I am preparing  the codebase to reuse the same parallel parquet row-group encoding/writing implementation for both TPCH and TPCDS.

## What changed

Extracted the TPCH parquet row-group encoding and writing path into a shared  module.

## Performance

I double checked that this did not change performance by running

```shell
cargo build --bin tpcgen-cli --release && \
hyperfine --runs 5 --prepare "rm -rf /tmp/output" "target/release/tpcgen-cli tpch parquet --scale-factor=10 --tables=lineitem --output-dir /tmp/output"
```

Results on main:

```
Benchmark 1: target/release/tpcgen-cli tpch parquet --scale-factor=10 --tables=lineitem --output-dir /tmp/output
  Time (mean ± σ):      3.830 s ±  0.940 s    [User: 33.783 s, System: 0.850 s]
  Range (min … max):    3.390 s …  5.511 s    5 runs
```

Results on this branch: 
```
Benchmark 1: target/release/tpcgen-cli tpch parquet --scale-factor=10 --tables=lineitem --output-dir /tmp/output
  Time (mean ± σ):      3.667 s ±  0.483 s    [User: 33.518 s, System: 0.880 s]
  Range (min … max):    3.424 s …  4.530 s    5 runs
```

PR #345, opened 2026-07-06
[arrow] Consolidate to use Arrow `RecordBatchReader` 

alamb:
## Summary

I noticed that both tpchgen_arrow and tpcdsgen_arrow redefined copy/paste versions of a `RecordBatchIterator` trait while working on https://github.com/clflushopt/tpchgen-rs/pull/342.

This is basically the same thing as  the`arrow` crate's existing `RecordBatchIterator`, so I propose to use that standard crate so it is easier to reuse the code.

The changes look large, but they are entirely mechanical -- I recommend reivewing using whitespace blind diff: https://github.com/clflushopt/tpchgen-rs/pull/345/changes?w=1


tpcdsgen-arrow/src/tables/call_center.rs
```diff
@@ -30,15 +32,15 @@ impl CallCenterArrow {
 }
 
 impl RecordBatchIterator for CallCenterArrow {
-    fn schema(&self) -> &SchemaRef {
-        &SCHEMA
+    fn schema(&self) -> SchemaRef {
```
alamb, line 39: This changes the API to match the arrow RecordBatchIterator API: 
1. return SchemaRef rather than &SchemaRef 
2. return `Result<RecordBatch>` rather than `RecordBatch`

Then that same change is applied to all the table generators

tpcdsgen-arrow/src/lib.rs
```diff
@@ -33,10 +31,7 @@ pub use tables::{
     TimeDimArrow, WarehouseArrow, WebPageArrow, WebReturnsArrow, WebSalesArrow, WebSiteArrow,
 };
 
-/// An iterator of Arrow [`RecordBatch`]es that also exposes its schema.
```
alamb, line 36: The whole point of this PR is to remove this (and the corresponding one in tpchgen-arrow

kevinjqliu:
rebased with help of codex

alamb:
shall I merge it?

Thank you @kevinjqliu

tpcgen-cli/src/parquet.rs
```diff
@@ -157,6 +158,7 @@ where
 
     // generate the data and send it to the tasks (via the sender channels)
     for batch in iter {
+        let batch = batch.unwrap();
```
kevinjqliu, line 161: small issue flagged by codex: 


> Since this PR switches the parquet path from the repo’s infallible `RecordBatchIterator` to Arrow’s `RecordBatchReader`, each reader item is now `Result<RecordBatch, ArrowError>`. This `unwrap()` would panic on a valid reader error even though `generate_parquet` returns `io::Result<()>`.
> 
> Can we propagate the batch error instead, with a small regression test using a reader that yields `Err`?

kevinjqliu (approved):
LGTM! 

codex flagged a minor issue around handling error from `RecordBatchReader`

tpcgen-cli/src/parquet.rs
alamb: Sure -- this is a good idea -- I will do so

Issue #334, opened 2026-07-06
[TPC-DS] Multi-core generation of parquet


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

Right now running `tpcgen tpcds parquet` uses 1 core. It should use them ALL

alamb:
Plan for adding TPCH-style parallel parquet generation to TPCDS:

1. Extract the generic parquet writer from `tpcgen-cli/src/tpch_cli/parquet.rs`.
   TPCH currently does the fast parquet path by encoding each planned row group into `ArrowColumnChunk`s concurrently, then appending those row groups through one `SerializedFileWriter`. This should become shared CLI code so TPCH and TPCDS do not duplicate the low-level parquet encoding/writing logic.

2. Keep TPCH behavior unchanged after extraction.
   TPCH should continue to use its existing `GenerationPlan`, row group sizing, compression, progress accounting, temp-file behavior, and writer path. This gives us a regression baseline before changing TPCDS behavior.

3. Add range-aware TPCDS Arrow generation.
   TPCDS Arrow iterators currently start at source row 1 and run to the table row count. Add constructors or adapters that can generate a source row range, for example `new_range(session, start_row, end_row)`. Internally this should use `skip_rows_until_starting_row_number(start_row)` where needed so generator RNG state remains correct.

4. Add a TPCDS parquet generation plan.
   Compute source row ranges from table row counts, estimated parquet bytes per row, and `--row-group-bytes`. Cap row groups at Parquet's limit. Each range becomes one `RecordBatchIterator`, mirroring TPCH's "one logical chunk per row group" model.

5. Wire TPCDS CLI options.
   TPCDS parquet already exposes `--row-group-bytes` but currently discards it. Make it real, and add `--num-threads` with the same default as TPCH so `tpcgen tpcds parquet` can use multiple cores for encode/write.

6. Replace the TPCDS `ArrowWriter` parquet path with the shared writer.
   The new TPCDS parquet path should create a temp `.inprogress` file, build row-group iterators from the TPCDS plan, call the shared parquet writer with compression/thread options, and rename on success.

7. Treat paired sales/returns tables carefully.
   `store_returns`, `catalog_returns`, and `web_returns` are generated from their corresponding sales generators. Their ranges must be based on the source sales rows, not output return row counts, to avoid missing or duplicating return rows.

8. Add focused tests.
[...]

PR #346, opened 2026-07-06
[tpcdsgen-arrow]: add source row range generation

alamb:
- Part of https://github.com/clflushopt/tpchgen-rs/issues/334 (Item 3)

The idea is that in order to generate parquet files in parallel, we need to initialize the TPCDS generators for different distinct parts of the input tables. 

Changes:
- Adds new_range(session, start_row, end_row) constructors for TPCDS Arrow table iterators.
- Add tests

### DIFF
--- a/tpcdsgen-arrow/src/tables/call_center.rs
+++ b/tpcdsgen-arrow/src/tables/call_center.rs
@@ -22,6 +22,10 @@ impl CallCenterArrow {
             batch_size: DEFAULT_BATCH_SIZE,
         }
     }
+    pub fn skip_rows_until_starting_row_number(&mut self, starting_row_number: i64) {
+        self.inner
+            .skip_rows_until_starting_row_number(starting_row_number);
+    }
 
     pub fn with_batch_size(mut self, batch_size: usize) -> Self {
         self.batch_size = batch_size;
--- a/tpcdsgen-arrow/tests/reparse.rs
+++ b/tpcdsgen-arrow/tests/reparse.rs
@@ -68,12 +68,15 @@ fn parse_dat<'a>(data: &'a [u8], schema: &'a SchemaRef) -> impl Iterator<Item = 
         .map(|batch| batch.expect("parse .tbl data into RecordBatch"))
 }
 
-/// Yields RecordBatches re-parsed from pipe-delimited output for the specified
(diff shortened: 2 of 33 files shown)


tpcdsgen-arrow/tests/reparse.rs
```diff
@@ -117,74 +120,120 @@ where
...
+        right.next().is_none(),
+        "right stream produced extra batches"
     );
 }
 
+/// Returns the source row to start skip tests from for the specified table.
+///
+/// Defaults to source row 100 and has special handling for slowly changing
+/// dimension (SCD) tables.
+fn skip_starting_row(table: Table, source_row_count: i64) -> i64 {
```
alamb, line 157: The tests to validate the skip results use the same basic plan as arrow in general -- it skips both a normal generator and the arrow generator and verifies the output is the same

tpcdsgen-arrow/src/tables/call_center.rs
```diff
@@ -22,6 +22,10 @@ impl CallCenterArrow {
             batch_size: DEFAULT_BATCH_SIZE,
         }
     }
+    pub fn skip_rows_until_starting_row_number(&mut self, starting_row_number: i64) {
```
alamb, line 25: This PR basically adds this method to the various arrow generators, fixes a few underlying generator issues, and adds a test

tpcdsgen-arrow/tests/reparse.rs
```diff
@@ -117,74 +120,120 @@ where
...
+    ) {
+        return max_skip_row;
+    }
+
+    // SCD tables reuse values from the previous source row for continuation
+    // records. Pick a new business-key row so a skipped generator does not need
+    // previous-row state initialized before the first generated row.
+    (1..=max_skip_row)
+        .rev()
+        .find(|row| row % 6 == 1)
```
kevinjqliu, line 171: 👍 , the `1` here means: this source row starts a fresh SCD business key that has only one historical version.

https://github.com/clflushopt/tpchgen-rs/blob/09d609d13b7b45a2aa06d2b86e5a4bfa29aacb1a/tpcdsgen/src/slowly_changing_dimension_utils.rs#L56

kevinjqliu (approved):
LGTM

PR #347, opened 2026-07-07
chore: remove generate chunks progress wrapper

kevinjqliu:
Similar to #344, remove the other internal wrapper

this is the last instance of the `_with_progress` pattern


alamb (approved):
🚀

PR #348, opened 2026-07-07
[tpcgen-cli] Propagate parquet reader errors

alamb:
## Summary

Propagate errors returned by Parquet RecordBatchReader inputs instead of panicking in the shared row-group encoder.

This follows up on the review comment from @kevinjqliu  in #345
- https://github.com/clflushopt/tpchgen-rs/pull/345#discussion_r3538620693

> Can we propagate the batch error instead, with a small regression test using a reader that yields Err?

Remove some unwrap and return errors

### DIFF
--- a/tpcgen-cli/src/parquet.rs
+++ b/tpcgen-cli/src/parquet.rs
@@ -76,7 +76,7 @@ where
                 encode_row_group(parquet_schema, writer_properties, schema, iter)
             })
             .await
-            .expect("Inner task panicked")
+            .map_err(|e| io::Error::other(format!("Inner task panicked: {e}")))?
         })
         .buffered(num_threads); // generate row groups in parallel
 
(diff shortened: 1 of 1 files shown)


tpcgen-cli/src/parquet.rs
```diff
@@ -76,7 +76,7 @@ where
                 encode_row_group(parquet_schema, writer_properties, schema, iter)
             })
             .await
-            .expect("Inner task panicked")
+            .map_err(|e| io::Error::other(format!("Inner task panicked: {e}")))?
```
alamb, line 79: The point of the PR is to remove the expect and unwraps.

I am not a huge fan of using io::Error everywhere --we should probably have a dedicated Error type

Maybe that is osmething we can do as a follow on PR

kevinjqliu (approved):
LGTM!

PR #349, opened 2026-07-07
[tpcgen-cli] Validate TPC-DS row group bytes with clap

alamb:
## Summary

Move TPC-DS parquet row group byte validation into ParquetArgs clap parsing instead of checking it inside ParquetArgs::run.

This follows up on the review comment from PR #341: https://github.com/clflushopt/tpchgen-rs/pull/341#discussion_r3533784008

## Details

The --row-group-bytes argument now parses directly as a usize using a dedicated clap value parser that rejects zero. ParquetArgs::run can pass the parsed value through without manual validation or conversion.

A CLI regression test asserts the full clap error message for --row-group-bytes 0.

## Validation

- cargo test

### DIFF
--- a/tpcgen-cli/src/tpcds_cli.rs
+++ b/tpcgen-cli/src/tpcds_cli.rs
@@ -12,6 +12,8 @@ pub mod parquet;
 
 type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
+const DEFAULT_TPCDS_PARQUET_ROW_GROUP_BYTES: usize = DEFAULT_PARQUET_ROW_GROUP_BYTES as usize;
+
 enum OutputFormat {
     Dat(dat::Dat),
     Csv(csv::Csv),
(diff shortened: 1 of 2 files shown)


tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -299,3 +291,12 @@ fn parse_delimiter(s: &str) -> std::result::Result<char, String> {
     }
     Ok(parsed)
 }
+
+fn parse_row_group_bytes(s: &str) -> std::result::Result<usize, String> {
+    let parsed = s.parse::<usize>().map_err(|e| e.to_string())?;
```
kevinjqliu, line 296: 👍 this should take care of both the negative numbers branch and the  "is too large" branch

kevinjqliu (approved):
LGTM! 

i updated the pr with the latest changes from main so CI can pass 😄

alamb:
Thanks @kevinjqliu

PR #351, opened 2026-07-07
refactor: simplify TPC-DS inventory DAT generation

kevinjqliu:
I was looking at the `dat.rs` code and noticed that `generate_inventory` only differs from `generate_simple` in its special row-count logic, which is already handled by [`Scaling::get_row_count`](https://github.com/clflushopt/tpchgen-rs/blob/main/tpcdsgen/src/config/scaling.rs#L46-L50).

We can remove this helper function and reuse `generate_simple` instead.


alamb (approved):
if the tests pass, ship it!

kevinjqliu:
🚢 ! thanks for the review

PR #352, opened 2026-07-07
[tpcgen-cli] Fix TPC-DS dat table selection

### LINKED ISSUE #301: [TPC-DS] Fix tpcgen DAT table selection for return tables

Follow-up from https://github.com/clflushopt/tpchgen-rs/pull/290

After wiring `tpcgen tpcds dat` to the existing TPC-DS DAT generator, table selection has a gap for the paired return tables.

These commands currently exit successfully but do not create the requested output file:

```shell
tpcgen tpcds dat --scale-factor 0 --tables store_returns --output-dir /tmp/tpcds
tpcgen tpcds dat --scale-factor 0 --tables catalog_returns --output-dir /tmp/tpcds
tpcgen tpcds dat --scale-factor 0 --tables web_returns --output-dir /tmp/tpcds
```

The underlying DAT generator treats `store_returns`, `catalog_returns`, and `web_returns` as no-ops because they are emitted as side effects of generating the corresponding sales tables:

- `store_sales` generates `store_sales.dat` and `store_returns.dat`
- `catalog_sales` generates `catalog_sales.dat` and `catalog_returns.dat`
- `web_sales` generates `web_sales.dat` and `web_returns.dat`

We should update the `tpcgen` TPC-DS DAT table-selection behavior so requesting one of the return tables either:

1. generates the requested return table output, or
2. rejects the selection with a clear error instead of succeeding with no output.

Ideally this should have CLI integration coverage for the three return-only selections.

alamb:
- Closes #301

## Summary
- Fix `tpcgen tpcds dat --tables <return_table>` so return-only selections generate the requested DAT output.
- Map DAT return-table selections to their paired sales generators, which emit both sales and returns files.
- Add DAT individual table-selection coverage to catch return-only no-output regressions.

## Root cause
The DAT writer emits `store_returns`, `catalog_returns`, and `web_returns` as side effects of the corresponding sales generators. Selecting a return table directly previously called a DAT match arm that returned `Ok(())`, so the command succeeded without creating the requested file.

## Validation
- `cargo fmt -p tpcgen-cli -- --check`
- `cargo test -p tpcgen-cli --test cli_integration tpcds::`

### DIFF
--- a/tpcgen-cli/tests/cli_integration/tpcds.rs
+++ b/tpcgen-cli/tests/cli_integration/tpcds.rs
@@ -7,6 +7,7 @@ use std::collections::BTreeSet;
 use std::fs;
 use std::fs::File;
 use tempfile::tempdir;
+use tpcdsgen::config::Table;
 
 /// Test that TPC-DS DAT generation is quiet unless logging is explicitly enabled.
 #[test]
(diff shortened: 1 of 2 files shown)


tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -343,6 +344,34 @@ fn test_tpcgen_cli_tpcds_dat_multiple_table_selection_command_forms() {
     }
 }
 
+/// Test each TPC-DS DAT table can be selected individually and creates output.
```
kevinjqliu, line 347: should we add tests for sales/returns tables behavior? 
- passing the "sales" table should generate both the "sales" and "returns" tables
- passing the "returns" table should also generate both the "sales" and "returns" tables
- passing both "sales" and "returns" tables should generate both tables, but only once (not twice)

The last one is a bug Ive seen previously. It takes twice as long and overwrites the original output on second generation

alamb: Yeah, I think the whole per table table generation thing needs to be rethought -- it is really strange to me that passing "sales"would generate two tables (I realize the internal generators are linked). 

What I think I will propose is to continue with this PR an file an issue to sort out multi-table generation

alamb: - Filed https://github.com/clflushopt/tpchgen-rs/issues/354 to track

kevinjqliu: sounds great, thanks!

kevinjqliu (approved):
LGTM!

kevinjqliu:
resolved merge conflict, will merge once CI is green

alamb:
Thanks @kevinjqliu

PR #353, opened 2026-07-08
refactor: share TPC-DS sales returns DAT generation

kevinjqliu:
This refactor is purely optional. I feel like it makes the code cleaner and easier to understand

## Summary
- keep table-specific sales/returns entry points for readability
- share the duplicated DAT loop through `generate_sales_and_returns`
- preserve one-pass paired generation for sales and optional returns rows

## Testing
- `cargo fmt --check`
- `cargo test -p tpcgen-cli test_tpcgen_cli_tpcds_dat_default_options_generate_all_outputs`
- `cd tpcdsgen && ./scripts/compare-all-tables.sh --quiet`

### DIFF
--- a/tpcgen-cli/src/tpcds_cli/dat.rs
+++ b/tpcgen-cli/src/tpcds_cli/dat.rs
@@ -326,32 +326,47 @@ fn generate_simple<G: RowGeneratorFactory>(
 }
 
 /// Generate store_sales and store_returns together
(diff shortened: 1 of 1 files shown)


alamb (approved):
This seems better to me than what is on main

In general I think we need to sort out optimizing the table generation . I filed a ticket to track
- https://github.com/clflushopt/tpchgen-rs/issues/354

Issue #355, opened 2026-07-08
Progress tracker integration followup


kevinjqliu:
Thanks for the review! I want to keep this PR strictly for the integration.

I'll note all these for follow up PR(s):
- Fix tpcds progress bar UI alignment (resolved in #356)
- change `get_row_count` to return a `u64` (rather than `i64`)
- incorporate the table name into `ProgressTracker`, to reduce number of parameters used for plumbing

_Originally posted by @kevinjqliu in https://github.com/clflushopt/tpchgen-rs/issues/350#issuecomment-4917845934_

kevinjqliu:
> incorporate the table name into ProgressTracker, to reduce number of parameters used for plumbing

a potential solution could be for `register` to return a `ProgressHandle` that gets plumbed through

alamb:
See also
- https://github.com/clflushopt/tpchgen-rs/issues/380

kevinjqliu:
a few more progress tracker related improvements:
- https://github.com/clflushopt/tpchgen-rs/pull/386
- https://github.com/clflushopt/tpchgen-rs/pull/393

we can call this done 😄

alamb:
woohoo

PR #356, opened 2026-07-08
Improve indicatif progress bar UI

kevinjqliu:
Part of #355

Improves the default progress bar implementation which uses `indicatif`'s `MultiProgress` 
- Renders all progress bar at 0% at start. This requires pre-registering all tables and calling the new `ProgressTracker::start()` 
- Progress labels and progress bars are now aligned 

You may notice that there are still some flickering, I believe this is due to parquet generation being **too fast** and calling `increment` too many times. We are already using a couple of suggestions recommended by `indicatif`, including:
- set `set_move_cursor(true);` to reduce flickering
- reduce refresh hz to 5 (default is 15)
To improve further, we need to coalesce the progress bar's `.inc` calls, which is a bigger change (we can explore as follow up)

### Parquet (updated gif after reducing refresh hz to 5)
https://github.com/user-attachments/assets/7b8fd104-d3a1-41c6-9e3d-b16a9cbb600c

### CSV
https://github.com/user-attachments/assets/86c392f4-1c5c-47b2-9b2f-1908288b9fe9

### Dat
https://github.com/user-attachments/assets/8175fd73-3be1-46b1-a2c5-db36bc8b5d79


alamb (commented):
Thank you @kevinjqliu 

For some reason it now goes across my entire scren  which is arguably better (though now there is more flashing)

Image

One thing that seems strange / a regression is that the labels don't appear until the bar starts going across

Image

While on main the label is present immediately
Image

alamb:
I tested locally again, and something still doesn't seem quite right 🤔 

The progress bar is not aligned during generation:

Screenshot 2026-07-08 at 5 30 56 PM

It only becomes aligned when all tables are done

Screenshot 2026-07-08 at 5 30 52 PM

tpcgen-cli/src/progress.rs
```diff
@@ -199,14 +212,12 @@ mod indicatif_impl {
         }
     }
 
-    fn bar_style() -> &'static ProgressStyle {
-        static STYLE: OnceLock<ProgressStyle> = OnceLock::new();
```
alamb, line 203: what is the implication and rationale for recreating the ProgressStyle multiple times rather than using a singleton?

alamb:
I may mess around with this a bit as well (I have noticed the progress bars of the rustup downloads which seem nicer and flash less)

kevinjqliu:
> I may mess around with this a bit as well (I have noticed the progress bars of the rustup downloads which seem nicer and flash less)

Feel free to take a look! 
I was exploring a way to deal with the sales tables since we cannot calculate a total during the planning phase. We have an estimate but the current interface is quite rigid, it requires the increments sums must add up to the pre-planned total (otherwise the progress bar looks incomplete).

kevinjqliu:
taking i took another look, the planning split for parquet makes the total/increment very simple! 
https://github.com/clflushopt/tpchgen-rs/pull/361/changes#diff-a9fabe7f29c8f5d32f3e86efb84101cb07f408b09c550180fa30c48f64ccf578R70

Once we do the same for dat/csv, the problem of some progress not reaching 100% goes away.

Let me update this PR to just fix the progress UI alignment issue

tpcgen-cli/src/progress.rs
kevinjqliu: made the change to use singleton now 😄

tpcgen-cli/src/progress.rs
```diff
@@ -158,55 +181,77 @@ mod indicatif_impl {
...
+                };
+                bars.values().cloned().collect::<Vec<_>>()
+            };
+
+            // Draw each registered bar at 0% before work starts.
+            for bar in bars {
+                bar.force_draw();
+            }
+            // Reduce flicker by moving the cursor instead of clearing lines once the bar set is stable.
+            self.multi.set_move_cursor(true);
```
kevinjqliu, line 216: helps reduce flickering https://docs.rs/indicatif/latest/indicatif/struct.MultiProgress.html#method.set_move_cursor

tpcgen-cli/src/tpcds_cli/csv.rs
```diff
@@ -33,6 +33,20 @@ impl Csv {
         }
     }
 
+    pub(super) fn register_table(
```
kevinjqliu, line 36: follows `parquet.rs`

kevinjqliu:
@alamb please take a look when you get time. I took a bunch of detours to find out the best way to render the progress bar. 
It seems that rendering all 25 bars at the beginning and redrawing on update is bottlenecked in `MultiProgress` and can cause cursor flickering, esp when all cores are used. 

Reducing refresh rate to 5hz (from 15hz) reduces the number of times the ui is redrawn. Setting `set_move_cursor(true)` also helps. A bigger change would be to coalesce the ProgressBar's `.inc` calls to further reduce redraw request pressure. 

Orrr we can render the progress bar one at a time 😄 (but i do like to see all the progress bars rendered at 0% up front)

kevinjqliu:
@alamb gentle ping in case you missed it. 

I have 2 follow up fix as well 
- coalescing the `increment` calls internally so it does not call `pb.inc()` every time. The number of updates, particularly from parquet generation, is what causes some of the flickering. 
- return ProgressHandle from register and pass that around, so we dont need to plumb through both ProgressTracker and the table name

tpcgen-cli/src/progress.rs
```diff
@@ -112,19 +122,25 @@ pub use indicatif_impl::IndicatifProgress;
...
     use super::ProgressTracker;
+    use indicatif::ProgressDrawTarget;
     use indicatif::{MultiProgress, ProgressBar, ProgressFinish, ProgressStyle};
     use std::collections::BTreeMap;
     use std::io::{self, Write};
     use std::sync::{OnceLock, RwLock};
 
+    const LABEL_WIDTH: usize = 22;
+    const BAR_WIDTH: usize = 18;
+    // 5 Hz redraws every 200 ms, keeping multi-bar updates responsive without repainting too often.
```
alamb, line 133: 👍

alamb (approved):
It worked great for me ❤️ 

Image

kevinjqliu:
> It worked great for me ❤️

awesome! It might be the vscode terminal thats acting up on my laptop.

> coalescing the increment calls internally so it does not call pb.inc() every time. The number of updates, particularly from parquet generation, is what causes some of the flickering.

I'll keep this one in the back pocket in case we need to further optimize the rendering

> return ProgressHandle from register and pass that around, so we dont need to plumb through both ProgressTracker and the table name

I'll push up a PR for this

PR #357, opened 2026-07-08
feat: Per-table row count scaling cache

clflushopt:
This was verbatim cherry-picked from #320 with a minor addition (test case) as suggested by @alamb


tpcdsgen/src/config/scaling.rs
```diff
@@ -279,6 +313,25 @@ mod tests {
         assert_eq!(customer_rows, 10000); // 100000 * 0.1
     }
 
+    #[test]
+    fn test_cached_row_counts_match_direct_computation() {
+        for scale in [0.1, 1.0, 10.0] {
```
alamb, line 318: this is a nice way to verify the implementations remain in sync

alamb (approved):
Looks good to me -- thanks @clflushopt

PR #359, opened 2026-07-08
feat: use fmt::Display buffering for Dat output

clflushopt:
This PR aligns `tpcdsgen` to use the same approach as `tpchgen` as we discussed in #320 this PR is kept small and only shows the change for two tables with the remaining tables in a separate PR.

### DIFF
--- a/tpcdsgen/src/output.rs
+++ b/tpcdsgen/src/output.rs
@@ -294,7 +294,7 @@ mod tests {
         let mut writer = DatWriter::new(&mut out, CompatMode::Trino);
         writer.write_display_row(&TestRow).unwrap();
         // Small rows stay buffered until an explicit flush.
-        assert!(writer.buffer().len() > 0);
+        assert!(!writer.buffer.is_empty());
         writer.flush().unwrap();
         // Ô (U+00D4) must come out as the single ISO-8859-1 byte 0xD4.
         assert_eq!(out, b"1|C\xD4TE|2.50|\n");
--- a/tpcdsgen/src/output.rs
+++ b/tpcdsgen/src/output.rs
@@ -20,9 +20,11 @@
 //!
 //! [`CompatWriter`] selects the right behavior based on [`CompatMode`].
 
+use std::fmt;
 use std::io::{self, Write};
 
 use crate::config::CompatMode;
+use crate::row::TableRow;
 
 /// Converts a UTF-8 string to ISO-8859-1 bytes.
 ///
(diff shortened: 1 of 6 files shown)


tpcdsgen/src/output.rs
```diff
@@ -136,6 +138,90 @@ impl<W: Write> Write for CompatWriter<W> {
...
+        if self.buffer.is_empty() {
+            return Ok(());
+        }
+        match self.compat_mode {
+            CompatMode::C => self.inner.write_all(&self.buffer)?,
+            CompatMode::Trino => {
+                if self.buffer.is_ascii() {
+                    self.inner.write_all(&self.buffer)?;
+                } else {
+                    let s = std::str::from_utf8(&self.buffer)
```
alamb, line 208: I was confused about the need for Dat writer - I think it is this line -- basically it is needs to transcode utf8 potentially

It would be nice to avoid the need for this extra layer somehow - but we can do it as a follow on PR perhaps

tpcdsgen/src/output.rs
```diff
@@ -136,6 +138,90 @@ impl<W: Write> Write for CompatWriter<W> {
...
+            compat_mode,
+            buffer: Vec::with_capacity(Self::FLUSH_THRESHOLD + 1024),
+        }
+    }
+
+    /// Write one row whose `Display` impl emits the DAT line (fields joined
+    /// by the separator, with a trailing separator); appends the newline.
+    pub fn write_display_row(&mut self, row: &impl fmt::Display) -> io::Result<()> {
+        writeln!(self.buffer, "{row}")?;
+        self.maybe_flush()
```
alamb, line 174: Since the outer code calls maybe_flush after writing each row, I think calling it again here is somewhat redundant

tpcgen-cli/src/tpcds_cli/dat.rs
```diff
@@ -469,27 +469,38 @@ fn create_output_file(path: &Path, output_options: &OutputOptions) -> Result<Fil
...
+        match row {
+            GeneratedRow::StoreSales(row) => writer.write_display_row(row)?,
+            GeneratedRow::StoreReturns(row) => writer.write_display_row(row)?,
+            row => writer.write_table_row(row, output_options.separator)?,
+        }
         return Ok(());
     }
 
+    let buffer = writer.buffer();
     for (i, value) in row.get_values().iter().enumerate() {
```
alamb, line 490: and is the idea to remove the `get_values()` call when we have ported all the code over to use this new pattern?

alamb (approved):
Makes sense to me @clflushopt  -- thanks

I had 2 questions, each of which could be done as af ollow on PR

tpcgen-cli/src/tpcds_cli/dat.rs
clflushopt: Yes !

tpcdsgen/src/output.rs
clflushopt: Yes I have a potential approach for this.

PR #361, opened 2026-07-08
Generate TPC-DS parquet tables concurrently (4.5%-35% faster)

alamb:
## Which issue does this PR close?

Follow on to #360 (part of #218)

## Rationale

After #360, TPC-DS parquet generation runs in parallel for each table, but still only processes one table at a time. This means that some cores sit idle while the trailing row groups of each table finish (and small tables run serially).

## What changes

Follow the TPCh model and convert all chunks of tables in one big stream so cores can pick up the next table when done with their current one

# Performance 
* Scale Factor 2: 20% faster

Command

I ran this on a GCP 8 core machine, testing both scale factor 1 and 10

| Branch | Scale Factor 1 (s) | Scale Factor 10 (s) 
|--------|--------|--------|
| main | 6.603 | 45.760 |
| this PR | 4.923 | 43.761 |

SF 1 speedup: 35%
SF10 speedup: 4.5%

The improvement is larger at smaller scale factor because the smaller tables take a larger proportion of the total time

```shell
hyperfine --runs 5 --prepare "rm -rf /tmp/output" "target/release/tpcgen-cli tpcds parquet --scale-factor=1  --output-dir /tmp/output" "target/release/tpcgen-cli tpcds parquet --scale-factor=10  --output-dir /tmp/output"
```

Full Results (MAIN)
```
Benchmark 1: target/release/tpcgen-cli tpcds parquet --scale-factor=1  --output-dir /tmp/output
  Time (mean ± σ):      6.603 s ±  0.166 s    [User: 31.624 s, System: 0.549 s]
  Range (min … max):    6.351 s …  6.768 s    5 runs

Benchmark 2: target/release/tpcgen-cli tpcds parquet --scale-factor=10  --output-dir /tmp/output
  Time (mean ± σ):     45.760 s ±  0.258 s    [User: 331.762 s, System: 3.823 s]
  Range (min … max):   45.384 s … 46.107 s    5 runs
```

Full results (BRANCH)
```
Benchmark 1: target/release/tpcgen-cli tpcds parquet --scale-factor=1  --output-dir /tmp/output
  Time (mean ± σ):      4.923 s ±  0.017 s    [User: 33.570 s, System: 0.861 s]
  Range (min … max):    4.907 s …  4.945 s    5 runs

Benchmark 2: target/release/tpcgen-cli tpcds parquet --scale-factor=10  --output-dir /tmp/output
  Time (mean ± σ):     43.761 s ±  0.117 s    [User: 328.248 s, System: 6.437 s]
[...]

PR #362, opened 2026-07-09
[tpcgen-cli]: extract WorkerQueue into a shared module

alamb:

Part of #218 (refactor split out of #361)

## Rationale

The TPC-H runner's `WorkerQueue` (schedule tasks with as many threads as they have chunks, within an overall thread budget) is exactly what TPC-DS parquet generation needs to generate multiple tables concurrently. Extract it so both can share it.

## What changes

- Move `WorkerQueue` from `tpch_cli/runner.rs` to a shared `worker_queue` module

### DIFF
--- a/tpcgen-cli/src/tpch_cli/runner.rs
+++ b/tpcgen-cli/src/tpch_cli/runner.rs
@@ -10,13 +10,13 @@ use crate::tpch_cli::parquet::generate_parquet;
 use crate::tpch_cli::tbl::*;
 use crate::tpch_cli::tbl::{LineItemTblSource, NationTblSource, RegionTblSource};
 use crate::tpch_cli::{OutputFormat, Table, WriterSink};
+use crate::worker_queue::WorkerQueue;
 use arrow::record_batch::RecordBatchReader;
 use log::{debug, info};
 use std::collections::BTreeMap;
 use std::io;
 use std::io::BufWriter;
 use std::sync::Arc;
-use tokio::task::{JoinError, JoinSet};
 use tpchgen::generators::{
     CustomerGenerator, LineItemGenerator, NationGenerator, OrderGenerator, PartGenerator,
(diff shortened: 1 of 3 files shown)


tpcgen-cli/src/tpch_cli/runner.rs
```diff
@@ -89,116 +89,22 @@ impl PlanRunner {
...
         // Do the actual work in parallel, using a worker queue
-        let mut worker_queue = WorkerQueue::new(num_threads, Arc::clone(&progress));
+        let mut worker_queue = WorkerQueue::new(num_threads);
         while let Some(plan) = plans.pop() {
-            worker_queue.schedule_plan(plan).await?;
+            debug!("scheduling plan {plan}");
+            let progress = Arc::clone(&progress);
+            worker_queue
+                .schedule(plan.chunk_count(), move |num_plan_threads| {
+                    run_plan(plan, num_plan_threads, progress)
```
alamb, line 98: run_plan is a tpch specific thing -- so to reuse the worker queue it needs to be passed in via callback rather than hard coded as below

tpcgen-cli/src/worker_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::Arc;
+
+    #[tokio::test]
+    async fn schedules_more_tasks_than_threads() {
```
alamb, line 110: these are some new tests that Codex added -- they seem like good coverage to me (esp the error one) but I could also be convinced to remove them too

alamb:
Any chance I can trouble you for a (quick) review @kevinjqliu  (I want to speed up parquet generation even more -- and this is a small pre-req)

kevinjqliu (approved):

kevinjqliu:
LGTM!

PR #363, opened 2026-07-09
cleanup: remove redundant generate_parquet function

alamb:
Removes the TPC-H parquet shim and lets the shared parquet writer derive the schema from the first RecordBatchReader directly. TPC-H and TPC-DS parquet callers now use the shared helper without an extra wrapper.


tpcgen-cli/src/tpch_cli/parquet.rs
```diff
@@ -1,47 +0,0 @@
...
-
-pub use crate::parquet::IntoSize;
-
-/// Converts a set of RecordBatchReaders into a Parquet file.
-///
-/// Uses num_threads to generate the data in parallel.
-///
-/// Note the input is an iterator of [`RecordBatchReader`]; the batches
-/// produced by each iterator are encoded as their own row group.
-pub async fn generate_parquet<W, I>(
```
alamb, line 18: the point is to remove this wrapper

alamb:
Any chance I can trouble you for a (quick) review @kevinjqliu  (I want to speed up parquet generation even more)

tpcgen-cli/src/parquet.rs
```diff
@@ -222,7 +221,6 @@ mod tests {
 
         generate_parquet(
             writer,
-            region_source().schema(),
```
kevinjqliu, line 225: 👍 , removed because derived from iter instead

kevinjqliu (approved):
LGTM

kevinjqliu:
can merge when CI is green, thanks!

PR #365, opened 2026-07-11
feat: Use `fmt::Display` for DAT override for all tables.

clflushopt:
This is a specific followup very mechanical change from #359 as of now this brings single core performance to about 4.9s and the attached profile captured via `samply` shows that roughly 55–60% of the remaining 4.9s is `fmt` machinery and about ~15% is real row generation and the rest is `memcpy`. 

I think single core performance can be brought to about ~2s (the napkin math suggests we can hit that) if we bypass the `core::fmt` machine and optimize for `itoa` style buffers with raw `write_str` calls where possible. But I don't see that as a net gain right now from a maintenance perspective since scaling out to multi-cores will net an even bigger performance benefit. So for now I propose we parallelize the generation machinery.

[tpcds-display-sf1.profile.json.gz](https://github.com/user-attachments/files/29929499/tpcds-display-sf1.profile.json.gz)


alamb:
I plan to review this hsortly

tpcdsgen/src/row/table_row.rs
```diff
@@ -28,6 +28,50 @@ pub(crate) fn dat_key(key: i64, is_null: bool) -> DatField<i64> {
     DatField((!is_null && key != -1).then_some(key))
 }
 
+/// DAT field from an already-computed optional value, for callers whose
```
alamb, line 31: As a follow on you could probably make these functions `impl From<Option<T>> for DatField<T>` to make it work more ergonomically with the rust ecosystem

alamb (approved):
I think it looks great

I ran performance tests and this looks like about the same as the current

Some follow ups for after this PR:
1. Apply the same patttern to the CSV path (with a custom CSV formatter)
1. Remove the `write_to` path is now dead (no longer called) 
2. I think we can remove all the OutputOptions too

tpcdsgen/src/row/table_row.rs
alamb: - Filed https://github.com/clflushopt/tpchgen-rs/issues/369

alamb:
Here is a follow on PR to remove the old formatting code
- https://github.com/clflushopt/tpchgen-rs/pull/385

PR #370, opened 2026-07-16
chore: remove standalone TPC-DS benchmark binary

alamb:

N/A

Removes the standalone TPC-DS `benchmark` binary and  the script whose only purpose was to build and run that binary.

## Rationale for this change

Now that we have tpcgen-cli wired up, we can do end to end benchmarking using the process described in [CONTRIBUTING.md#benchmarking](https://github.com/clflushopt/tpchgen-rs/blob/main/CONTRIBUTING.md#benchmarking), so the standalone benchmark binary is redundant.

## Are these changes tested?

Covered by existing CI


tpcgen-cli/src/bin/benchmark.rs
```diff
@@ -1,769 +0,0 @@
...
-            let num_rows = session.get_scaling().get_row_count(Table::CallCenter);
-            if !no_output {
-                let file = File::create(output_dir.join("call_center.dat"))?;
-                let mut writer = Iso8859Writer::new(BufWriter::new(file));
-                for row_number in 1..=num_rows {
-                    let result =
-                        gen.generate_row_and_child_rows(row_number, session, None, None)?;
-                    gen.consume_remaining_seeds_for_row();
-                    for row in result.get_rows() {
-                        row.write_to(&mut writer, '|')?;
```
alamb, line 74: this is also the last real use of write_to 

Once this PR is merged, I think we can remove write_to and all the supporting code

kevinjqliu (approved):
LGTM!

alamb:
Here is a small follow on (to remove some more code)
- https://github.com/clflushopt/tpchgen-rs/pull/375

PR #371, opened 2026-07-16
refactor(tpcds): remove unused OutputOptions from DAT output

alamb:

Follow-up to #365

## Rationale for this change

`OutputOptions` is left over from the Java implementation but the CLI never exposed any of them

Remove the old OutputOptions

## Are these changes tested?

By CI

No


alamb:
I think this will make it simpler to update CSV generation

kevinjqliu (approved):
LGTM!

PR #377, opened 2026-07-17
fix TPC-DS conformance script validation

kevinjqliu:
## Summary

- return failure for invalid TPC-DS script options while keeping `--help` successful
- require all 25 non-empty C fixtures before treating a fixture set as complete

## Why

- Invalid options previously printed an error but exited successfully, so a typo such as `--ful` instead of `--full` could skip conformance checks without failing the caller. 
- The C fixture readiness check inspected only five representative files, even though its result determines whether downloading can be skipped. Checking every expected table ensures an incomplete fixture set is repaired before comparison starts.

## Validation

- `bash -n` on both scripts
- `shellcheck` on both scripts
- focused exit-code and partial/full fixture-set probes

### DIFF
--- a/tpcgen-cli/scripts/tpcds/generate-fixtures.sh
+++ b/tpcgen-cli/scripts/tpcds/generate-fixtures.sh
@@ -75,7 +75,6 @@ Examples:
 
 See scripts/README.md for the full conformance-testing workflow.
 EOF
-    exit 0
 }
 
 # Colors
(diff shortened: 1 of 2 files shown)


tpcgen-cli/scripts/tpcds/generate-fixtures.sh
```diff
@@ -292,18 +291,18 @@ check_c_prerequisites() {
     return 0
 }
 
-# Sanity check the extracted fixtures.
-# At minimum, a handful of expected tables must be present and non-empty.
+# Verify that the extracted fixture set is complete.
 verify_c_fixtures() {
     local fixture_dir=$1
-    local required=(store_sales.dat catalog_sales.dat web_sales.dat reason.dat call_center.dat)
```
kevinjqliu, line 299: Already generating `ALL_TABLES`, so we can also verify that all the tables are present

https://github.com/clflushopt/tpchgen-rs/blob/edc7cfe52d52f6f6a26b477826e4e7c266b910ad/tpcgen-cli/scripts/tpcds/generate-fixtures.sh#L496

alamb:
this is a nice one

tpcgen-cli/scripts/tpcds/generate-fixtures.sh
alamb: Makes sense

alamb (approved):
love it -- thanks @kevinjqliu

PR #378, opened 2026-07-17
feat: introduce `ProgressHandle` for easier plumbing 

kevinjqliu:
Part of #355

Previously, we had to plumb both `progress: Arc<dyn ProgressTracker>` and `table_name: &'static str` all the way down the stack so we can do `progress.increment(table_name, 1)` to increment the progress tracker. 

This is now all handled in the new `ProgressHandle`, and we just need to create it per table and plumb it through.

### DIFF
--- a/tpcgen-cli/src/parquet.rs
+++ b/tpcgen-cli/src/parquet.rs
@@ -15,7 +15,7 @@ use std::io::Write;
 use std::sync::Arc;
 use tokio::sync::mpsc::{Receiver, Sender};
 
-use crate::progress::ProgressTracker;
+use crate::progress::ProgressHandle;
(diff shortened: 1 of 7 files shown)


kevinjqliu:
I also explored a more involved follow-up:
- Change `ProgressTracker::register(self: Arc<Self>, ...) -> ProgressHandle`
- Remove `increment` from `ProgressTracker`

That would be an API change, with `ProgressTracker` owning the run-level lifecycle and `ProgressHandle` owning increments. I’m not sure yet whether it actually makes the API cleaner enough to justify the extra registration and handle-storage changes.

clflushopt (approved):
neat !

PR #379, opened 2026-07-18
refactor: let `ProgressHandle` handle increments

kevinjqliu:
Part of #355

## Summary

- `ProgressTracker::register` now returns an item-bound `ProgressHandle`.
- `ProgressHandle::increment` handles item-level progress updates.

### DIFF
--- a/tpcgen-cli/src/parquet.rs
+++ b/tpcgen-cli/src/parquet.rs
@@ -200,8 +200,10 @@ mod tests {
     }
 
     impl ProgressTracker for CountingProgress {
-        fn increment(&self, _item: &str, row_groups: u64) {
-            self.increments.fetch_add(row_groups, Ordering::Relaxed);
+        fn register(self: Arc<Self>, _item: &str, _total_units: u64) -> ProgressHandle {
+            ProgressHandle::new(move |row_groups| {
+                self.increments.fetch_add(row_groups, Ordering::Relaxed);
+            })
         }
     }
 
@@ -217,7 +219,7 @@ mod tests {
 
         let tracker = Arc::new(CountingProgress::default());
         let progress: Arc<dyn ProgressTracker> = tracker.clone();
-        let progress = ProgressHandle::new(progress, "ignored");
+        let progress = progress.register("ignored", 2);
 
         generate_parquet(
             writer,
--- a/tpcgen-cli/src/progress.rs
+++ b/tpcgen-cli/src/progress.rs
@@ -2,24 +2,25 @@
 //!
 //! # Overview
 //!
-//! [`ProgressTracker`] is a small, dyn-compatible trait that receives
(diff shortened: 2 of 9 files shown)


tpcgen-cli/src/parquet.rs
kevinjqliu, line 1: this is just changing the test example

tpcgen-cli/src/tpcds_cli/dat.rs
```diff
@@ -38,6 +39,16 @@ pub(super) struct Dat {
     output_dir: PathBuf,
 }
 
+#[derive(Debug)]
+pub(super) enum DatProgress {
```
kevinjqliu, line 43: this is somewhat hacky to deal with sales/returns tables

tpcgen-cli/src/tpch_cli/generate.rs
kevinjqliu, line 1: this is just changing the test example

tpcgen-cli/src/tpch_cli/generator.rs
kevinjqliu, line 1: this is just changing the test example

clflushopt (approved):

tpcgen-cli/src/tpcds_cli/dat.rs
alamb: yeah, i need to figure out if there is some cleaner way to separate the generation of those tables

alamb:
I tested this using 
```shell
cargo run --release --bin tpcgen-cli tpcds dat --scale-factor=1
```

Something doesn't seem quite right with the progress bars now (they don't make it all the way to the right for some tables): 

Screenshot 2026-07-20 at 7 17 05 AM


output


However, I double checked and this same behavior happens on `main` so I think we can handle it as a follow on PR

Edit: filed
- https://github.com/clflushopt/tpchgen-rs/issues/380

tpcgen-cli/src/progress.rs
```diff
@@ -103,25 +99,38 @@ pub trait ProgressTracker: Send + Sync + fmt::Debug {
...
-/// the responsibility of the tracker owner.
-#[derive(Clone, Debug)]
+/// Handles are created by [`ProgressTracker::register`] and directly own the
+/// item-specific increment behavior. Run-level lifecycle operations remain the
+/// responsibility of the tracker owner.
+#[derive(Clone)]
 pub struct ProgressHandle {
-    tracker: Arc<dyn ProgressTracker>,
-    item: &'static str,
+    increment: Arc<dyn Fn(u64) + Send + Sync>,
```
alamb, line 107: this is a nice refactor It hink

alamb (approved):

alamb:
Thanks @kevinjqliu and @clflushopt  -- the refactor certainly looks nice

Issue #354, opened 2026-07-08
[TPC-DS] Avoid redundantly generating tables in CLI


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

Follow up from 
- https://github.com/clflushopt/tpchgen-rs/issues/301

> Specifying a parent/child parent (store_sales & store_returns) will generate store_returns twice, second run overwrites the first. Not a correctness bug but definitely unnecessary

@kevinjqliu  tried to simplify this too
- https://github.com/clflushopt/tpchgen-rs/pull/353/changes

Related comment https://github.com/clflushopt/tpchgen-rs/pull/352#discussion_r3540466553 from @kevinjqliu :

> should we add tests for sales/returns tables behavior? 
> - passing the "sales" table should generate both the "sales" and "returns" tables
> - passing the "returns" table should also generate both the "sales" and "returns" tables
> - passing both "sales" and "returns" tables should generate both tables, but only once (not twice)
> 
> The last one is a bug Ive seen previously. It takes twice as long and overwrites the original output on second generation

alamb:
This also means that specifying one table like `--table call_center`  actually generates 4 files:

```
File: /var/folders/1l/tg68jc6550gg8xqf1hr4mlwr0000gn/T/.tmpvYz2b5/call_center.dat
File: /var/folders/1l/tg68jc6550gg8xqf1hr4mlwr0000gn/T/.tmpjVHHPQ/catalog_page.dat
File: /var/folders/1l/tg68jc6550gg8xqf1hr4mlwr0000gn/T/.tmpSjCoSP/catalog_returns.dat
File: /var/folders/1l/tg68jc6550gg8xqf1hr4mlwr0000gn/T/.tmpSjCoSP/catalog_sales.dat


Selecting catalog_returns should not create file for catalog_sales: "/var/folders/1l/tg68jc6550gg8xqf1hr4mlwr0000gn/T/.tmpSjCoSP/catalog_sales.dat"
thread 'tpcds::test_tpcgen_cli_tpcds_dat_individual_table_selection_outputs_requested_table' panicked at tpcgen-cli/tests/cli_integration/tpcds.rs:384:13:
Selecting catalog_returns should not create file for catalog_sales: "/var/folders/1l/tg68jc6550gg8xqf1hr4mlwr0000gn/T/.tmpSjCoSP/catalog_sales.dat"
```

[...]

PR #382, opened 2026-07-20
feat: list TPC-DS tables and descriptions in `--tables` help

alamb:

## Rationale for this change

`tpcgen-cli tpch --help` lists the possible values for `--tables`:

```
  -T, --tables <TABLES>
          Which tables to generate (default: all)

          Possible values:
          - region:   Region table (alias: r)
          - nation:   Nation table (alias: n)
          ...
```

but the TPC-DS equivalent showed nothing, leaving users to guess the table names:

```
  -T, --tables <TABLES>
          Which tables to generate (default: all)
```

Add a `TableValueParser` for TPC-DS mirroring the TPC-H one, so `tpcgen-cli tpcds --help` lists every selectable table with a one line description:

```
  -T, --tables <TABLES>
          Which tables to generate (default: all)

          Possible values:
          - call_center:            Call center dimension
          - catalog_page:           Catalog page dimension
          - catalog_returns:        Catalog returns fact
          - catalog_sales:          Catalog sales fact
          ...
          - web_site:               Web site dimension
          - dbgen_version:          Metadata about the generator run
```

The parser also gives `--tables` a typed `Vec<Table>` rather than `Vec<String>`, removing a parse/re-parse round trip. TPC-DS table names have no single letter aliases (unlike TPC-H), so no alias text is shown.

## Are these changes tested?

Yes:

Yes, improved `--help` output. No change to accepted values or error messages

### DIFF
--- a/tpcgen-cli/src/tpcds_cli.rs
+++ b/tpcgen-cli/src/tpcds_cli.rs
@@ -4,6 +4,7 @@ use crate::logging::configure_logging;
 use crate::progress::IndicatifProgress;
 use crate::progress::{no_op_progress_tracker, ProgressTracker};
 use crate::tpch_cli::{Compression, DEFAULT_PARQUET_ROW_GROUP_BYTES};
+use clap::builder::TypedValueParser;
 use clap::{ArgAction, Args, Subcommand};
 use std::io;
 #[cfg(feature = "indicatif-progress")]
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):
LGTM! 

Perhaps as a follow up, we can add a fix for https://github.com/clflushopt/tpchgen-rs/issues/354 in the clap parser

For example, this will generate the same file 3 times. 
```
--tables reason,reason,reason
```

Issue #339, opened 2026-07-06
[TPC-DS] Multi-core generation of csv/dat


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

Similarly to https://github.com/clflushopt/tpchgen-rs/issues/334, right now running `tpcgen tpcds <csv|dat> ` uses 1 core. It should use them ALL

clflushopt:
Working on this after #365 is merged

alamb:
I made a PR to extract out CSV generation and have it follow the same pattern as the tpch (and now DAT) output
- https://github.com/clflushopt/tpchgen-rs/pull/373

Once we merge this I suspect we can probably just reuse the same generator trait from tpch and get parallelization

PR #383, opened 2026-07-20
test: cover CSV in the TPC-DS reparse tests

alamb:

- part of https://github.com/clflushopt/tpchgen-rs/issues/339

## Rationale for this change

I want to ensure our CSV generation code is good;

We rely on comparing the dat output directly to other generators, and then we compare the other formats (arrow/csv) to that dat output.

Change the CSV checks to reparse CSV to Arrow and compare directly against the generated batches ensures correctness by transitivity

## Are these changes tested?

Yes

No.

### DIFF
--- a/tpcdsgen-arrow/tests/reparse.rs
+++ b/tpcdsgen-arrow/tests/reparse.rs
@@ -1,6 +1,11 @@
 //! Verifies correctness of the tpcdsgen-arrow generators by reparsing the
-//! canonical pipe-delimited .dat format and comparing against the directly
-//! generated Arrow RecordBatches.
+//! textual output formats (pipe-delimited `.dat` and CSV) and comparing against
+//! the directly generated Arrow RecordBatches.
+//!
+//! This also serves as a transitive test for csv correctness: we compare the
+//! DAT output to the original C and Trino generators, and verify it is the same
+//! as arrow. Thus, if CSV is the same as arrow, it too is the same as the
+//! original generators.
 //!
 //! Strategy:
 //! - drive the tpcdsgen RowGenerator to produce rows for each table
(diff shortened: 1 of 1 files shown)


tpcdsgen-arrow/tests/reparse.rs
```diff
@@ -90,19 +147,16 @@ where
     let mut source_row = starting_row_number;
     std::iter::from_fn(move || {
         let mut data = Vec::new();
+        format.write_header(table, &mut data);
```
alamb, line 151: have to account for csv headers

tpcdsgen-arrow/tests/reparse.rs
```diff
@@ -48,33 +55,83 @@ fn test_row_count(table: Table) -> i64 {
...
+        let null_re = regex::Regex::new("^$").unwrap();
+        let builder =
+            arrow_csv::reader::ReaderBuilder::new(Arc::clone(schema)).with_null_regex(null_re);
+        let builder = match self {
+            Format::Dat => builder
+                .with_delimiter(DAT_SEPARATOR as u8)
+                .with_header(false),
+            Format::Csv => builder
+                .with_delimiter(CSV_SEPARATOR as u8)
+                .with_header(true),
```
kevinjqliu:
```suggestion
                .with_header(true)
                .with_header_validation(true),
```
nit, codex flagged this:
Could we also validate the CSV header? DAT has no header, so its branch can remain unchanged.
https://arrow.apache.org/rust/arrow_csv/reader/struct.ReaderBuilder.html#method.with_header_validation

kevinjqliu (approved):
LGTM!

PR #386, opened 2026-07-20
Fix incomplete and noisy progress rendering

### LINKED ISSUE #380: Progress bars don't seem to make it all the way to the right
I tested this using 
```shell
cargo run --release --bin tpcgen-cli tpcds dat --scale-factor=1
```

Something doesn't seem quite right with the progress bars now (they don't make it all the way to the right for some tables): 

Screenshot 2026-07-20 at 7 17 05 AM

output

_Originally posted by @alamb in https://github.com/clflushopt/tpchgen-rs/issues/379#issuecomment-5021666769_

kevinjqliu:
- Fixes #380

This addresses incomplete progress bars in two cases:

1. Exact-count progress bars could finish without a final visible redraw. The indicatif-backed progress handle now explicitly finishes the bar when its counter reaches the registered total.

2. Some generators can produce a different number of increments than the registered total. This happens for paths such as paired sales/returns table generation, where the registered total may not exactly match the emitted rows. `ProgressHandle::complete` lets those generators explicitly finish the bar after successful generation.

This also reduces cursor flicker from high-frequency progress updates. Some paths report progress from tight row loops or multiple threads, so calling `ProgressBar::inc` for every `ProgressHandle::increment` can repeatedly enter indicatif’s render path. The indicatif progress backend now batches row increments and forwards them to indicatif at most every 200 ms.

With throttled `ProgressBar::inc` calls, the previous display workarounds are no longer needed, so this removes:

- the custom 5 Hz `MultiProgress` draw target
- `set_move_cursor(true)`

Tested
```
cargo run -p tpcgen-cli --bin tpcgen-cli -- tpcds dat --scale-factor=1 -o `mktemp -d`
cargo run -p tpcgen-cli --bin tpcgen-cli -- tpcds csv --scale-factor=1 -o `mktemp -d`
cargo run -p tpcgen-cli --bin tpcgen-cli -- tpcds parquet --scale-factor=1 -o `mktemp -d`
```

#### DAT
https://github.com/user-attachments/assets/e9de7d1b-eba0-4683-b212-738af6b8aa76

#### CSV
https://github.com/user-attachments/assets/0122ef91-455d-4826-9af3-8990e3dcfe8b

#### Parquet
https://github.com/user-attachments/assets/42432f3b-3993-4b9a-8d96-ca86a3d04adf

### DIFF
--- a/tpcgen-cli/src/progress.rs
+++ b/tpcgen-cli/src/progress.rs
@@ -174,15 +174,16 @@ pub use indicatif_impl::IndicatifProgress;
 #[cfg(feature = "indicatif-progress")]
 mod indicatif_impl {
     use super::{ProgressHandle, ProgressTracker};
+    #[cfg(test)]
     use indicatif::ProgressDrawTarget;
     use indicatif::{MultiProgress, ProgressBar, ProgressFinish, ProgressStyle};
     use std::io::{self, Write};
     use std::sync::{Arc, Mutex, MutexGuard, OnceLock};
+    use std::time::{Duration, Instant};
 
     const LABEL_WIDTH: usize = 22;
     const BAR_WIDTH: usize = 18;
-    // 5 Hz redraws every 200 ms, keeping multi-bar updates responsive without repainting too often.
-    const PROGRESS_REFRESH_HZ: u8 = 5;
+    const PROGRESS_FLUSH_INTERVAL: Duration = Duration::from_millis(200);
     const PROGRESS_CHARS: &str = "=>-";
 
     /// Default [`ProgressTracker`] implementation backed by
@@ -203,9 +204,7 @@ mod indicatif_impl {
         /// [`ProgressTracker::register`].
         pub fn new() -> Self {
             Self {
-                multi: MultiProgress::with_draw_target(ProgressDrawTarget::stderr_with_hz(
-                    PROGRESS_REFRESH_HZ,
-                )),
+                multi: MultiProgress::new(),
                 bars: Mutex::new(Vec::new()),
             }
         }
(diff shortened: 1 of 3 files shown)


tpcgen-cli/src/progress.rs
```diff
@@ -225,15 +242,32 @@ mod indicatif_impl {
...
+                }
+            };
+
+            let on_complete = move || {
+                // Finish when the caller reports successful item completion, even if
+                // the counter has not reached its registered total.
+                complete_bar.finish_using_style();
+            };
+
+            ProgressHandle::new_with_complete(on_increment, on_complete)
```
kevinjqliu: this is the main logic!

alamb:
Thanks @kevinjqliu  -- I am sorry but I tried this out and something still doesn't seem right. The bars go all the way over but once store_sales starts, then it looks like my terminal starts repainting like mad (the cursor flashes and things slow way down)

output

I reproduced via

```shell
cargo build --release
./target/release/tpcgen-cli tpcds dat --scale-factor=1
```

Maybe it is continually setting the progress even when there is no change or somethign?

tpcgen-cli/src/progress.rs
```diff
@@ -256,6 +274,113 @@ mod indicatif_impl {
...
+                total,
+                throttle: Mutex::new(ThrottleState {
+                    pending: 0,
+                    next_flush,
+                }),
+                clock: Arc::new(now),
+            }
+        }
+
+        fn increment(&self, units: u64) {
```
kevinjqliu, line 326: Added this time-based per progress bar throttling which calls `pb.inc` at most every 200ms.

This is better than `MultiProgress`'s refresh hz since it is per progress bar and helps when `increment` is called many times by multiple threads.

kevinjqliu:
@alamb thats the same cursor flickering issue i saw before!! Setting `MultiProgress` to refresh every 5 hz (instead of default 15) and adding `set_move_cursor(true)` was the previous workaround. It looks like the number of `increment` calls for `store_sales` is really really high and it locks up the rendering. 

I added a per progress bar throttling. I tested it locally and it seem to resolve the issue (we can even get rid of the previous workarounds!) 

Please give it a try when you get a chance. I sincerely hope this is the last progress bar related issue 😭

tpcgen-cli/src/progress.rs
alamb: maybe worth a small comment about the design

alamb (commented):
Well, the flickering has stopped. But now the progress bars seem to go back to being incomplete. 

Image

I feel like we may be missing something more fundamental here -- like could we  use a more coarse grained progress counter (rather than row count) or something 🤔

kevinjqliu:
> Well, the flickering has stopped. But now the progress bars seem to go back to being incomplete.

thats really odd, we call [`complete()`](https://github.com/clflushopt/tpchgen-rs/pull/386/changes#diff-61ca30bfa0bf52ae343d193d8941fb13a34c44b162a62ac5c6ad46e2af18703aR201) directly so that line should hit 100% as soon as the generation for the table is complete.

I see youre using the compiled binary, could it be outdated? Could you try
```
cargo run -p tpcgen-cli --bin tpcgen-cli -- tpcds dat --scale-factor=1 -o `mktemp -d`
```

> I feel like we may be missing something more fundamental here -- like could we use a more coarse grained progress counter (rather than row count) or something 🤔

I agree! I like the way parquet generation break up the work in chunks (in the worker queue), and then we just count those chunks in progress tracker

alamb:
```shell
cargo run --release -p tpcgen-cli --bin tpcgen-cli -- tpcds dat --scale-factor=1 -o `mktemp -d`
```

It ran well!

> I agree! I like the way parquet generation break up the work in chunks (in the worker queue), and then we just count those chunks in progress tracker

I hope to do this shortly

alamb (approved):
This has been hanging out for too long -- let's get this in!

PR #387, opened 2026-07-21
chore: Add tpcdsgen-arrow CI coverage, use `cargo nextest`

alamb:
Our current CI setup runs a job per-crate tests, which means 
1. we recompile the same code several many times (not so efficient)
2. We don't necessarily run tests with newly added crates 

In fact, we weren't running tpcdsgen-arrow (52 tests) in the CI job at all under the per-crate setup

I think it is both more efficient and less error prone to run   CI job for all the tests using `nextest`

The [new test job ](https://github.com/clflushopt/tpchgen-rs/actions/runs/29831308013/job/88637108641?pr=387)now takes 7 minutes for all the tests, where the 4 old jobs each took around 4 minutes each. 

However, several of the package build tests take over 10 minutes anyways so this doesn't increase the end to end latency of CI jobs


.github/workflows/rust.yml
```diff
@@ -25,59 +25,28 @@ jobs:
...
-      - uses: actions/checkout@v7
-      - name: All Tests (tpcdsgen)
-        run: cargo test -p tpcdsgen
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest@0.9.140
+      # nextest runs lib, bins and test targets by default (doc tests are
+      # not supported by nextest and are covered by the job above)
+      - name: All Tests (workspace)
+        run: cargo nextest run --workspace
```
alamb, line 49: by using `--workspace` here we'll automatically pick up all newly added tests

.github/workflows/rust.yml
```diff
@@ -92,6 +61,7 @@ jobs:
 
   # Build wheels + sdist (but don’t upload artifacts)
   build-packages:
+    needs: lint
```
alamb, line 64: Don't bother building packages until we make sure clippy passes -- that will avoid burning CI on jobs that will need to be rerun if clippy/lint fails

kevinjqliu (approved):
LGTM! I also noticed we did not add `tpcdsgen-arrow`, glad `--workspace` is taking care of it and any new ones

PR #389, opened 2026-07-21
perf: parallelize TPC-DS DAT generation

alamb:
**NOT READY FOR REVIEW**

Part of #339 (and #218)

## NOTE: Being Broken Up

To make this PR easier to review, I am breaking it up into the following 5 PRs:

| # | PR | Scope | ~Size | Depends on |
|---|---|---|---|---|
| https://github.com/datafusion-contrib/tpcgen-rs/pull/459 | Move shared generation plumbing to the crate root | `generate` (plus a new `generate_file`), `sink` (`WriterSink`), `statistics`; TPC-H import updates | +110 / −75, nearly all moves | — |
| https://github.com/datafusion-contrib/tpcgen-rs/pull/460 | Move `RowIter` to `tpcdsgen`; add `GeneratedRow::table()` | `tpcdsgen`, `tpcdsgen-arrow` only | +118 / −72 | — |
| https://github.com/datafusion-contrib/tpcgen-rs/pull/462 | Each TPC-DS returns table is its own output | **Behaviour change.** Still serial. Deletes `generate_sales_and_returns`, `row_generator_tables`, `TableProgress::Paired`; updates the conformance script and tests | +80 / −270 | 2 |
| #461 | Extract `plan_tables` / `run_plans` into `tpcds_cli::runner` | Parquet only, pure refactor; renames `row_group_count` → `chunk_count` | +150 / −90 | — |
| https://github.com/datafusion-contrib/tpcgen-rs/pull/463 | Parallel DAT + CSV | `ChunkFormat` + per-format byte estimates, `RowFormat` / `RowSource`, rewrite `dat.rs` / `csv.rs`, `--num-threads`, chunk-based progress | +650 / −400 | 1, 2, 3, 4 |

PRs 1, 2 and 4 are pure refactors with no behaviour change and should review quickly. PR 3
is the one that needs real scrutiny, and it is ~80 added lines.

PRs 1 and 2 are open as #459 and #460; both target `main` and can be reviewed
independently of each other.

# Which issue does this PR close?

Closes part of #339 — `tpcgen-cli tpcds dat` used a single core.

# Rationale for this change

DAT generation walked one table at a time with a single row generator writing
straight to its output file. Everything needed to do better already exists in
this crate: TPC-H (and TPC-DS Parquet) generate chunks in parallel and hand
them to a single ordered writer.

# What changes are included in this PR?

Generate DAT with that same mechanism:

* Each table is split into chunks of source rows (`TpcdsGenerationPlan`, now
  shared between the Parquet and DAT outputs with per-format estimates of
  output bytes per source row).
* Chunks are formatted into in-memory buffers on separate threads
  (`generate_in_chunks`), and one writer task appends them to the file in order.
* Tables are generated concurrently within the overall thread budget
  (`WorkerQueue`), largest first.

The pieces TPC-H and TPC-DS now share moved to the crate root of `tpcgen-cli`:

* `generate` — `Source`, `Sink`, `generate_in_chunks`
* `sink` — `WriterSink`
* `statistics` — `WriteStatistics`

`RowIter` moved from `tpcdsgen-arrow` to `tpcdsgen` (generating a range of
source rows is not Arrow specific), and `GeneratedRow` gained a `table()`
accessor so each output file can keep just its own rows.

CSV is deliberately untouched — that will follow once #373 lands.

# Are these changes tested?

* All 25 tables at scale factor 1 match the committed
  `tests/fixtures/tpcds/scale-1-trino/MD5SUMS` byte for byte (except
  `dbgen_version.dat`, which records the command line and time of the run).
* New unit tests: chunked generation concatenates to exactly the unchunked
  output, sales/returns rows are split correctly across chunks, empty ranges
  produce empty output.
* New plan tests for the DAT chunk sizing.

# Are there any user-facing changes?

* `tpcgen-cli tpcds dat` accepts `--num-threads`, defaulting to the number of CPUs.
* Selecting only a returns table (e.g. `--tables store_returns`) now writes only
  that file. Previously it also wrote the sales table, because returns were
  emitted as a side effect of the sales generator. This matches the Parquet output.

# Performance

Scale factor 1, all tables, 10 core laptop:

| | wall |
|---|---|
| before | 7.2s |
| after  | 1.8s |
| after, `--num-threads 1` | 8.3s |

Single threaded generation costs ~15% more because the sales tables are now
walked twice, once for the sales rows and once for the returns rows — the same
trade the Parquet output already makes.

### DIFF
--- a/tpcdsgen-arrow/src/lib.rs
+++ b/tpcdsgen-arrow/src/lib.rs
@@ -18,10 +18,7 @@
 pub mod conversions;
 mod tables;
 
-use std::collections::VecDeque;
-
-use tpcdsgen::config::Session;
-use tpcdsgen::row::{GeneratedRow, RowGenerator};
+pub(crate) use tpcdsgen::row::RowIter;
 
 pub use tables::{
     CallCenterArrow, CatalogPageArrow, CatalogReturnsArrow, CatalogSalesArrow,
@@ -33,72 +30,3 @@ pub use tables::{
 
 /// Default number of rows per [`RecordBatch`](arrow::array::RecordBatch).
 pub const DEFAULT_BATCH_SIZE: usize = 8_000;
-
-/// Adapts a [`RowGenerator`] into a streaming [`Iterator`] of [`GeneratedRow`]s.
-///
-/// Handles both simple generators (one row per call, `should_end_row` always
-/// true) and paired fact-table generators (multiple calls per source row,
-/// `should_end_row` signals when to advance the row counter).
-pub(crate) struct RowIter<G: RowGenerator> {
-    generator: G,
-    session: Session,
-    current_row: i64,
-    row_count: i64,
-    pending: VecDeque<GeneratedRow>,
-}
-
-impl<G: RowGenerator> RowIter<G> {
-    pub(crate) fn new(generator: G, session: Session, row_count: i64) -> Self {
-        Self {
-            generator,
-            session,
-            current_row: 1,
-            row_count,
-            pending: VecDeque::new(),
-        }
-    }
-
-    pub(crate) fn skip_rows_until_starting_row_number(&mut self, starting_row_number: i64) {
-        self.generator
-            .skip_rows_until_starting_row_number(starting_row_number);
-        self.current_row = starting_row_number;
-        self.pending.clear();
-    }
-
-    /// Restrict generation to source rows
-    /// `starting_row_number..=ending_row_number` (1-based, inclusive).
-    ///
-    /// The ending row number is clamped to the table's row count.
-    pub(crate) fn set_source_row_range(
-        &mut self,
-        starting_row_number: i64,
-        ending_row_number: i64,
-    ) {
-        self.skip_rows_until_starting_row_number(starting_row_number);
-        self.row_count = self.row_count.min(ending_row_number);
-    }
-}
-
-impl<G: RowGenerator> Iterator for RowIter<G> {
-    type Item = GeneratedRow;
-
-    fn next(&mut self) -> Option<GeneratedRow> {
-        while self.pending.is_empty() {
-            if self.current_row > self.row_count {
-                return None;
-            }
-            let result = self
-                .generator
-                .generate_row_and_child_rows(self.current_row, &self.session, None, None)
-                .expect("row gen");
-            for row in result.get_rows() {
-                self.pending.push_back(row.clone());
-            }
-            if result.should_end_row() {
-                self.generator.consume_remaining_seeds_for_row();
-                self.current_row += 1;
-            }
-        }
-        self.pending.pop_front()
-    }
-}
--- a/tpcdsgen/src/row/generated_row.rs
+++ b/tpcdsgen/src/row/generated_row.rs
@@ -19,6 +19,7 @@
 //!
 //! See ISSUE-004 for details.
 
+use crate::config::Table;
 use crate::row::{
     CallCenterRow, CatalogPageRow, CatalogReturnsRow, CatalogSalesRow, CustomerAddressRow,
     CustomerDemographicsRow, CustomerRow, DateDimRow, DbgenVersionRow, HouseholdDemographicsRow,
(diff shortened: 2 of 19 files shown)


alamb:
This PR has been split into a reviewable stack: #459, #460, #461, #462, #463. It stays open as the integration reference until those land.

<details>
<summary>Full context (for picking this work up cold)</summary>

## Where things stand

| PR | Branch | Base | Content |
|---|---|---|---|
| #459 | `alamb/tpcds-1-shared-plumbing` | `main` | Move `generate` / `sink` / `statistics` to the `tpcgen-cli` crate root, add `generate_file` |
| #460 | `alamb/tpcds-2-row-iter` | `main` | Move `RowIter` to `tpcdsgen`, add `GeneratedRow::table()` |
| #461 | `alamb/tpcds-4-tpcds-runner` | `main` | Extract `plan_tables` / `run_plans` into `tpcds_cli::runner` |
| #462 | `alamb/tpcds-3-returns-own-output` | `main` | Each TPC-DS returns table becomes its own output (**behaviour change**) |
| #463 | `alamb/tpcds-5-parallel-dat-csv` | `main` | Parallel DAT + CSV generation |

All five are CI-green. #462 contains #460's commits; #463 contains all four others. Both target `main` so that the TPC-DS conformance workflow runs (see "CI gotcha" below). Once the others land, #463 should be rebased on `main` so its diff collapses to just its own last commit.

Dependencies: #459, #460 and #461 touch disjoint code and are independent of each other. #462 needs #460 (for `GeneratedRow::table()`). #463 needs all four.

## What this PR originally did, and what changed

The original branch parallelized DAT only, on a base ~49 commits behind `main`. Merging `main` brought in `--parts`/`--part` support (#406) and the TPC-DS CSV output, both of which had been built on a *serial* row-generator driver. Reconciling them meant reworking the TPC-DS output layer rather than resolving conflicts textually, and extending the parallelism to CSV.

## The constraint that drives the ordering

A parallel `Source` produces **one buffer for one file**, and `generate_in_chunks` writes to a single sink. So the pre-existing behaviour where a sales generator *also* writes its returns file cannot survive parallelization. That is why #462 (returns tables as independent outputs) must land before #463, and why it is the only user-visible behaviour change in the series.

## Design

All three TPC-DS outputs now share one path, which is also the TPC-H path:

- `tpcds_cli::runner::plan_tables` — plan each `(table, part)` into chunks of source rows, register one progress bar per table shared across its parts, drop empty `--parts` chunks
- `tpcds_cli::runner::run_plans` — schedule on a `WorkerQueue`, largest plan first
- `crate::generate::generate_in_chunks` / `generate_file` — generate chunks in parallel, single writer appends in order, temp file renamed into place

DAT and CSV differ only in how a chunk of rows becomes a buffer, captured by the `RowFormat` trait (`EXTENSION`, `write_header`, `write_rows`). `TpcdsGenerationPlan` takes a `ChunkFormat` so one planner sizes chunks for Parquet, DAT and CSV.

Notable decisions:

- **`generate_rows` in #462 does not use `RowIter`.** `RowIter` hides the source-row boundary and that PR's progress ticks per source row, so it keeps an explicit loop mirroring `RowIter`'s stepping. #463 deletes that loop and uses `RowIter` with per-chunk progress.
- **Progress units.** #462 counts source rows; #463 switches DAT/CSV to chunks, matching Parquet, and deletes `register_table`.
- **`Scaling::get_row_count(StoreReturns)` is `0`.** The old `store_returns` progress bar was therefore sized to zero and never moved. #462 fixes this by sizing bars from source rows.
- **Text chunk size is a constant** (`DEFAULT_TEXT_CHUNK_SIZE_BYTES`, 8 MiB), not a flag, unlike Parquet's `--row-group-bytes`.
- **DAT/CSV byte estimates** in `plan.rs` were measured from scale-factor-1 output. Unlike the Parquet estimates they do not vary with scale factor, because the field values come from the same distributions at every scale.

## Behaviour changes

1. Selecting one of a sales/returns pair writes **only** that table's file. Previously DAT and CSV wrote both, because returns were emitted as a side effect of the sales generator. This matches Parquet. `tpcgen-cli/scripts/tpcds/compare-table.sh` relied on the old behaviour and was updated to ask for the table it wants.
2. `tpcgen-cli tpcds dat` and `tpcgen-cli tpcds csv` accept `--num-threads`, defaulting to the number of CPUs.

## Performance (scale factor 1, 10-core laptop)

| Output | Before | After | Single-threaded |
|---|---|---|---|
| DAT | 7.8s | 1.0s | 8.6s |
| CSV | 8.4s | 1.2s | 8.7s |

#462 on its own is ~20% *slower* (DAT 7.0s → 8.5s, CSV 7.2s → 8.4s), because a sales table and its returns table each walk the generator once. This is the same trade the Parquet output already makes. It could be removed entirely if `generate_in_chunks` supported one source fanning out to two sinks — a real design change, deliberately out of scope here.

## How each PR was verified

Beyond the workspace test suite, clippy (`-D warnings`), `cargo doc` (`-D warnings`), `cargo fmt --check`, `--no-default-features` and `typos`:

```shell
# conformance against the committed scale-1 fixtures
./tpcgen-cli/scripts/tpcds/compare-all-tables.sh --compat trino   # 24/24
./tpcgen-cli/scripts/tpcds/compare-all-tables.sh --compat c       # 24/24

# byte-for-byte output comparison against the branch's base
cargo build --release --bin tpcgen-cli
./target/release/tpcgen-cli tpcds <dat|csv|parquet> -s 1 -o /tmp/new --no-progress -q
./target/release/tpcgen-cli tpcds <dat|csv|parquet> -s 1 --parts 4 \
  --tables customer_demographics,reason,store_returns -o /tmp/newparts --no-progress -q
# ...then rebuild from the base commit and `diff -rq`
```

`dbgen_version` always differs: it records the command line and the time of the run. Everything else must match exactly.

At the tip of #463, all three formats are byte-identical to the stacked base at scale factor 1, whole-table and with `--parts 4`.

## CI gotcha

`.github/workflows/tpcds-conformance.yml` has `pull_request: branches: [ main, master ]`, so the TPC-DS conformance suites **only run when a PR's base branch is `main`**. `tpch-conformance.yml` and `rust.yml` use a bare `pull_request:` and run on any base. A stacked TPC-DS PR therefore silently loses TPC-DS conformance coverage — which is why #462 and #463 target `main` rather than their logical parents. Worth fixing separately by dropping the branch filter (and adding `workflow_dispatch:`, which that workflow lacks).

</details>

PR #393, opened 2026-07-26
feat: show ETA in progress bars

### LINKED ISSUE #358: [FEATURE] Progress Bars show MB/s and ETA
I  ran `rustup update` and I saw that in addition to progress bars, it also reported Bytes/second and an ETA

It would be pretty amazing if we could do something similar for the tpchgen

Image

kevinjqliu:
- Closes #358 

## Description

Adds ETA estimate to CLI progress bars. Throughput (Bytes/second ) is harder to implement since progress is currently reported in chunks, rows, or Parquet row groups rather than bytes.

https://github.com/user-attachments/assets/d64fae3b-c285-4f3a-9bbd-507855266393

### DIFF
--- a/tpcgen-cli/src/progress.rs
+++ b/tpcgen-cli/src/progress.rs
@@ -385,15 +385,19 @@ mod indicatif_impl {
         static STYLE: OnceLock<ProgressStyle> = OnceLock::new();
         STYLE
             .get_or_init(|| {
-                let template = format!(
-                    "{{msg:!{LABEL_WIDTH}}} [{{bar:{BAR_WIDTH}.cyan/blue}}] ({{percent:>3}}%)"
-                );
+                let template = bar_template();
                 ProgressStyle::default_bar()
                     .template(&template)
                     .expect("progress bar template is valid")
                     .progress_chars(PROGRESS_CHARS)
             })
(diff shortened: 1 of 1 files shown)


tpcgen-cli/src/progress.rs
```diff
@@ -396,6 +394,12 @@ mod indicatif_impl {
             .clone()
     }
 
+    fn bar_template() -> String {
+        format!(
+            "{{msg:!{LABEL_WIDTH}}} [{{bar:{BAR_WIDTH}.cyan/blue}}] ({{percent:>3}}%) ETA {{eta}}"
```
kevinjqliu, line 399: same as before, with ` ETA {{eta}}` added to the end

alamb (approved):
I tried it out
```shell
cargo run --release --bin tpcgen-cli tpch parquet --scale-factor=10
```

And

```shell
cargo run --release --bin tpcgen-cli tpcds parquet --scale-factor=100
```

it works pretty well -- it is sort of strange that yo only get ETAs for each table once it starts (the other ETAs stay zero) but I think it makes sense

kevinjqliu:
thanks for the review @alamb

PR #399, opened 2026-08-19
Improve progress ETA display

kevinjqliu:
Context: https://github.com/clflushopt/tpchgen-rs/pull/393#pullrequestreview-4971884561

Shows `ETA --` until a table starts and `done` on completion. 

Also fixes a bug where a table’s ETA included time spent waiting for earlier tables. The ETA timer now starts when the table writes its first row, showing `ETA --` until then and done after completion.

## Demo
tpcgen-eta-progress


clflushopt (approved):
nice catch ! I was actually debugging this yesterday 😛

PR #400, opened 2026-08-19
feat: improve testing docs

### LINKED ISSUE #384: [TPC-DS] Document testing /conformance strategy
- Part of https://github.com/clflushopt/tpchgen-rs/issues/218

For verification of tpcds output, we rely on comparing the `dat` output directly to the g generator as well as the trino java generator.

We then we compare the other formats (arrow/parquet and csv) to that dat output

This should be clearly documented if it isn't already, so both we and readers have confidence in the correctness of these crates

clflushopt:
- closes  #384

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,50 +1,147 @@
+# Testing
 
-This repository has extensive tests to ensure correctness. On every checkin,
-CI compares the generators' output against the original reference
-implementations (`dbgen` for TPC-H, the Trino Java port and `dsdgen` for
-TPC-DS) using MD5 checksums committed to the repo.
+This repository checks its generators against the original reference
+implementations: `dbgen` for TPC-H, and both the C `dsdgen` and the Trino Java
+port for TPC-DS. Comparisons are byte-for-byte rather than statistical.
 
-The conformance suites live next to the unified CLI crate:
+Only the text formats (`.tbl` for TPC-H, `.dat` for TPC-DS) can be compared to a
+reference implementation directly, because those are the only formats the
+reference implementations produce. Arrow, CSV, and Parquet are verified against
+that text output instead, so their correctness follows from its.
 
-- TPC-H: [tpcgen-cli/scripts/tpch/](tpcgen-cli/scripts/tpch/) —
-  `compare-all-tables.sh` checks output against
-  `tpcgen-cli/tests/fixtures/tpch/scale-N/MD5SUMS` (generated from C `dbgen`
-  by `generate-fixtures.sh`, which requires docker or podman).
-- TPC-DS: [tpcgen-cli/scripts/tpcds/](tpcgen-cli/scripts/tpcds/) — see its
-  [README](tpcgen-cli/scripts/tpcds/README.md) for the Trino (`--compat
-  trino`) and C `dsdgen` (`--compat c`) suites.
+## The chain of trust
+
+```
+  C dbgen (TPC-H)     --.
+  C dsdgen (TPC-DS)   --+--> .tbl / .dat --> Arrow --+--> CSV
+  Trino Java (TPC-DS) --'                            '--> Parquet
+```
+
+### Text output against the reference implementations
+
+`tpcgen-cli tpch tbl` and `tpcgen-cli tpcds dat` are compared to reference data
+produced by the original generators. Comparison is by MD5 of the whole file,
+with an optional row-level `diff` when a hash does not match. The expected
+hashes live in `tpcgen-cli/tests/fixtures/` and are committed to the repo, so
+the common case needs neither a container runtime nor a reference build.
+
+The suites live next to the unified CLI crate:
+
+- TPC-H: [tpcgen-cli/scripts/tpch/](tpcgen-cli/scripts/tpch/). `compare-all-tables.sh`
+  checks output against `tpcgen-cli/tests/fixtures/tpch/scale-N/MD5SUMS`,
+  generated from C `dbgen` by `generate-fixtures.sh` (which needs docker or
+  podman).
+- TPC-DS: [tpcgen-cli/scripts/tpcds/](tpcgen-cli/scripts/tpcds/). Two reference
+  implementations, selected with `--compat trino` (the Java port the Rust code
+  was derived from) and `--compat c` (the TPC-supplied C `dsdgen`, whose
+  pre-generated data comes from [alamb/tpcds-data]). See the suite's
+  [README](tpcgen-cli/scripts/tpcds/README.md) for details, and
+  [tpcdsgen/BUGS.md](tpcdsgen/BUGS.md) for the Java bugs that `--compat c`
+  corrects.
 
 ```sh
 ./tpcgen-cli/scripts/tpch/compare-all-tables.sh --scale 1
 ./tpcgen-cli/scripts/tpcds/compare-all-tables.sh --scale 1
+./tpcgen-cli/scripts/tpcds/compare-all-tables.sh --scale 1 --compat c
 ```
 
-Conformance runs in two tiers:
+TPC-H has a second, finer-grained check at the library level:
+`tpchgen/tests/integration_tests.rs` compares each generator row by row against
+gzipped reference tables checked into `tpchgen/data/` at scale factors 0.001 and
+0.01, so a failure points at a single record rather than a file hash.
 
-- **Every PR** (`tpch-conformance.yml`, `tpcds-conformance.yml`): fast
-  MD5-only checks against the committed MD5SUMS — this proves output is
-  byte-identical to what the recorded reference hashes describe, with no
-  container runtime or reference build needed.
-- **Every merge to main** (`full-conformance.yml`): the slow pass. It
-  regenerates reference data from the reference implementations themselves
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+```
+  C dbgen (TPC-H)     --.
+  C dsdgen (TPC-DS)   --+--> .tbl / .dat --> Arrow --+--> CSV
+  Trino Java (TPC-DS) --'                            '--> Parquet
+```
+
+### Text output against the reference implementations
+
+`tpcgen-cli tpch tbl` and `tpcgen-cli tpcds dat` are compared to reference data
+produced by the original generators. Comparison is by MD5 of the whole file,
```
alamb: It might help to add a link directly here to the "compared to reference data" to the script that does the comparison.

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+with an optional row-level `diff` when a hash does not match. The expected
```
alamb: I think the row level diff and the rest of this sentence is sort of a implementation detail that we could omit here when explaining the testing high level testing strategy. We could remove it and make this section easier to understand

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+hashes live in `tpcgen-cli/tests/fixtures/` and are committed to the repo, so
+the common case needs neither a container runtime nor a reference build.
+
+The suites live next to the unified CLI crate:
+
+- TPC-H: [tpcgen-cli/scripts/tpch/](tpcgen-cli/scripts/tpch/). `compare-all-tables.sh`
```
alamb: I think the details of how this works are too much for this doc and make it harder to understand

I would suggest just leaving the link here for people to follow if they want more details

So something like

Please see the comparison scripts for more details:
* TPC-H: [tpcgen-cli/scripts/tpch/](tpcgen-cli/scripts/tpch/)
 ...


And if there are more details needed on how they work (like generate-fixtures) we should update the readmes in those directories

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
-- **Every merge to main** (`full-conformance.yml`): the slow pass. It
-  regenerates reference data from the reference implementations themselves
-  (C `dbgen` in docker, the C `dsdgen` published dataset, and a fresh Maven
-  build of the Trino Java port), byte-for-byte diffs our output against it,
-  and fails if the committed MD5SUMS have drifted from the living reference.
-
-`tpcgen-cli tpch` generates **exactly** the same bytes as the original
-`dbgen` program. You can verify this for yourself by using `shasum`, for
-example:
+TPC-H has a second, finer-grained check at the library level:
```
alamb, line 51: Would be good to link directly to `tpchgen/tests/integration_tests.rs`

Also I think " so a failure points at a single record rather than a file hash." is too detailed so could be removed and not reduce understanding

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+that text output instead, so their correctness follows from its.
 
-- TPC-H: [tpcgen-cli/scripts/tpch/](tpcgen-cli/scripts/tpch/) —
-  `compare-all-tables.sh` checks output against
-  `tpcgen-cli/tests/fixtures/tpch/scale-N/MD5SUMS` (generated from C `dbgen`
-  by `generate-fixtures.sh`, which requires docker or podman).
-- TPC-DS: [tpcgen-cli/scripts/tpcds/](tpcgen-cli/scripts/tpcds/) — see its
-  [README](tpcgen-cli/scripts/tpcds/README.md) for the Trino (`--compat
-  trino`) and C `dsdgen` (`--compat c`) suites.
+## The chain of trust
```
alamb, line 12: I know what this chain of trust means but I think it would be easier to understand if:
1. We added a sentence summarizing it. Something like "The validation strategy relies on a chain of trust -- the .tbl/.dat output is checked byte for byte against reference implementation, and then csv and parquet output is checked against the .tbl/.dat  format"
2. Aligned the headings of the next section to something in the diagram (so it is clear how those sections map back to the diagram)

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
-`dbgen` program. You can verify this for yourself by using `shasum`, for
-example:
+TPC-H has a second, finer-grained check at the library level:
+`tpchgen/tests/integration_tests.rs` compares each generator row by row against
+gzipped reference tables checked into `tpchgen/data/` at scale factors 0.001 and
+0.01, so a failure points at a single record rather than a file hash.
+
+### Arrow against the text output
+
+`tpchgen-arrow/tests/reparse.rs` and `tpcdsgen-arrow/tests/reparse.rs` write
```
alamb: I think here it would be easier to summarize more briefly and refer to the scripts for implementation details

So like
1. Add links to `tpchgen-arrow/tests/reparse.rs` and `tpcdsgen-arrow/tests/reparse.rs`
2. Say something like "these tests verify the libraries CSV output by reading both the CSV and DAT using the Arrow CSV reader and comparing them for (exact) equality. Since the text output is
known to match the reference implementations, matching it transitively proves the
Arrow output is correct as well.


The details about covering all 24 tables is obvious and I think the details about coverage could be shorted ("also covers chunked generation")

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
-
-`tpcgen-cli tpch` generates **exactly** the same bytes as the original
-`dbgen` program. You can verify this for yourself by using `shasum`, for
-example:
+TPC-H has a second, finer-grained check at the library level:
+`tpchgen/tests/integration_tests.rs` compares each generator row by row against
+gzipped reference tables checked into `tpchgen/data/` at scale factors 0.001 and
+0.01, so a failure points at a single record rather than a file hash.
+
+### Arrow against the text output
```
alamb: Maybe we can change this heading to focus on what is being validated (in this case the CSV output?) -- so like

### Validating CSV Output

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+from the start, which is what parallel generation relies on.
+
+### CSV against Arrow
+
+The same reparse tests run a second time over CSV output, using
+`tpcdsgen::csv::{csv_header, GeneratedRowCsv}` for TPC-DS and the `*Csv` row
+wrappers in `tpchgen::csv` for TPC-H. These are the types
+`tpcgen-cli/src/tpcds_cli/csv.rs` and its TPC-H counterpart use to write real
+CSV files, so the test exercises the shipped formatting code rather than a
+parallel implementation of it.
```
alamb, line 72: I don't understand this section (or how it is different than above) -- it seems like it is still discussing CSV validation (though in this case when integrated into the CLI). Maybe you could just list this as a third type of test for CSV (and again, link to the actual test file, rather than try to describe it in detail here)

TESTING.md
alamb: I may be confused though 🤔

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+
+### Parquet against Arrow
+
+Parquet is written from the Arrow batches, so the CLI integration tests in
```
alamb: Similarly here -- I think it would be much clearer if:
1. The section focused n "validating the parquet output"
2. Explained brieflty how it is done (generate Arrow output directly, and then compare against reading parquet files. Since we verified Arrow above, transitively this then verifies the parquet output)
3. Provide a link to the test implementation rather than try to summarize it here.

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+  gives the same data as a single pass (`test_tpcgen_cli_tpcds_parquet_matches_single_pass_generation`).
+  `store_returns` is generated from the `store_sales` generator, so this also
+  proves no return rows are lost or duplicated at range boundaries.
+- `--num-threads 1` and `--num-threads 4` produce byte-identical files.
+- The Arrow schema survives the round trip, including types with no exact
+  Parquet equivalent such as the `Time32(Second)` column in `dbgen_version`.
+
+## Conformance in CI
+
+Conformance runs in two tiers.
```
alamb: I hink we could summarize this section with "All the conformance tests described above runs on every CI runs" and then point to the two ci scripts. I think enumerating how it works in prose here is likely to get out of date an is confusing

TESTING.md
```diff
@@ -1,50 +1,147 @@
...
+  no workflow currently checks them. CI verifies TPC-DS at scale factor 1
+  (both compat modes) and, in the full pass, scale factor 2 for C.
+- The reparse tests run in Trino compat mode only, since that is
+  `Session::default()`. The `--compat c` corrections are covered at the `.dat`
+  level but not through the Arrow, CSV and Parquet paths.
+- The step from text back to Arrow uses the `arrow-csv` reader as its parser, so
+  a bug in that reader that exactly mirrored a bug in our Arrow generators would
+  not be caught.
+
+## Verifying it yourself
```
alamb, line 117: Could we also add a section here for tpcds data generation?

alamb (approved):
Thanks @clflushopt  -- I left a bunch of comments, but I think this PR is already better than what we have today on main so we could also merge it in and iterate on it as a follow on

alamb (approved):
looks much better -- thank you @clflushopt

PR #405, opened 2026-08-28
Support a `--column-encoding` flag to specify per-column encoding for parquet files

### LINKED ISSUE #237: [FEATURE] Per-column encoding
This request follows the pattern of #229 but addresses encoding rather than compression.

I'd suggest adding a `--column-encoding` option that accepts key-value pairs for column names and their desired encodings.
```
tpchgen-cli parquet --column-encoding="l_comment=DELTA_LENGTH_BYTE_ARRAY,l_shipinstruct=DELTA_LENGTH_BYTE_ARRAY"
```

Add a new cli argument to parquet `--column-encoding` and pass it to the `parquet-rs` writer: https://docs.rs/parquet/latest/parquet/file/properties/struct.WriterPropertiesBuilder.html#method.set_column_encoding

**Additional context**
xref #229 
xref #173 (Instead of calling the cli arg `--parquet-column-encoding`, we'll call it `--column-encoding` for example)

mroeschke:
closes #237

Adds `--column-encoding` flag to the tpchgen-cli CLI with `parquet` to support a comma delimited series of per-column encodings when generating parquet data.

Note: I am not super proficient in Rust so this PR was heavily agent generated.

Also I'm a colleague of @Matt711, who raised this issue, and we're very interested in using this feature

### DIFF
--- a/tpcgen-cli/src/parquet.rs
+++ b/tpcgen-cli/src/parquet.rs
@@ -7,9 +7,10 @@ use log::debug;
 use parquet::arrow::arrow_writer::{compute_leaves, ArrowColumnChunk};
 use parquet::arrow::{add_encoded_arrow_schema_to_metadata, ArrowSchemaConverter};
 use parquet::basic::{Compression, Encoding};
-use parquet::file::properties::WriterProperties;
+use parquet::file::properties::{WriterProperties, WriterPropertiesBuilder};
 use parquet::file::writer::SerializedFileWriter;
 use parquet::schema::types::{ColumnPath, SchemaDescPtr};
+use std::collections::HashSet;
 use std::io;
 use std::io::Write;
 use std::str::FromStr;
@@ -35,6 +36,41 @@ pub(crate) fn parse_column_encoding_pair(s: &str) -> Result<(String, Encoding), 
     }
     let encoding = Encoding::from_str(encoding).map_err(|e| e.to_string())?;
     Ok((name.to_string(), encoding))
+}
+
+fn apply_column_encodings(
+    mut builder: WriterPropertiesBuilder,
+    schema: &arrow::datatypes::Schema,
+    encodings: &[(String, Encoding)],
+) -> io::Result<WriterPropertiesBuilder> {
+    let names: HashSet<&str> = schema.fields().iter().map(|f| f.name().as_str()).collect();
+    for (col, enc) in encodings {
+        if !names.contains(col.as_str()) {
+            return Err(io::Error::other(format!(
+                "unknown column '{col}' for --column-encoding"
+            )));
+        }
+        match enc {
+            Encoding::PLAIN_DICTIONARY | Encoding::RLE_DICTIONARY => {
+                return Err(io::Error::other(format!(
+                    "encoding {enc} cannot be set with --column-encoding; dictionary encoding is the writer default. Use a non-dictionary encoding such as PLAIN or DELTA_LENGTH_BYTE_ARRAY"
+                )));
+            }
+            #[allow(deprecated)]
+            Encoding::BIT_PACKED => {
+                return Err(io::Error::other(
+                    "encoding BIT_PACKED is not supported for Parquet writing",
+                ));
+            }
+            _ => {
+                let path = ColumnPath::from(col.as_str());
+                builder = builder
+                    .set_column_encoding(path.clone(), *enc)
+                    .set_column_dictionary_enabled(path, false);
+            }
+        }
+    }
+    Ok(builder)
 }
 
 /// Converts a set of RecordBatchReaders into a Parquet file.
@@ -68,11 +104,8 @@ where
 
     // Compute the parquet schema
     let mut builder = WriterProperties::builder().set_compression(parquet_compression);
-    for (col, enc) in column_encodings.into_iter().flatten() {
-        let path = ColumnPath::from(col.as_str());
-        builder = builder
-            .set_column_encoding(path.clone(), *enc)
-            .set_column_dictionary_enabled(path, false);
+    if let Some(encodings) = column_encodings {
+        builder = apply_column_encodings(builder, schema.as_ref(), encodings)?;
     }
     let mut writer_properties = builder.build();
     // Embed the Arrow schema in the Parquet metadata (as ArrowWriter does) so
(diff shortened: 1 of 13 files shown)


tpcgen-cli/src/tpch_cli/output_plan.rs
```diff
@@ -35,7 +35,21 @@ impl Display for OutputLocation {
     }
 }
 
-/// Describes an output partition (file) that will be generated
```
clflushopt, line 38: I believe this comment to have been accidentally deleted it should be at position L:52 as it documents `OutputPlan` struct.

tpcgen-cli/src/tpch_cli/output_plan.rs
```diff
@@ -35,7 +35,21 @@ impl Display for OutputLocation {
...
+/// Parquet writer settings applied when generating parquet output.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ParquetWriteOptions {
```
clflushopt: I suggest it be renamed to `ParquetWriterOptions`

tpcgen-cli/src/tpch_cli/output_plan.rs
```diff
@@ -35,7 +35,21 @@ impl Display for OutputLocation {
...
+    pub compression: Compression,
+    pub column_encodings: Vec<(String, Encoding)>,
```
clflushopt: Having `pub column_encodings: Option<Vec<(String, Encoding)>>` should simplify the code here as these are optional.

tpcgen-cli/src/tpch_cli/output_plan.rs
```diff
@@ -35,7 +35,21 @@ impl Display for OutputLocation {
...
+}
+
+impl Default for ParquetWriteOptions {
+    fn default() -> Self {
+        Self {
+            compression: Compression::SNAPPY,
+            column_encodings: Vec::new(),
```
clflushopt: With the type change to `Option` this can be `None`

tpcgen-cli/src/tpch_cli/output_plan.rs
```diff
@@ -97,7 +110,11 @@ impl OutputPlan {
 
     /// Return the parquet compression level for this partition
     pub fn parquet_compression(&self) -> Compression {
-        self.parquet_compression
+        self.parquet.compression
+    }
+
+    pub fn parquet_column_encodings(&self) -> &[(String, Encoding)] {
```
clflushopt: The change I recommended above to `Option<...>` will make this signature much more idiomatic as the presence of a reference to an empty array requires doing checks such as `.len() == 0` for the default option (which is the more likely to be used).

clflushopt (changes requested):
I've left a few comments that I believe must be addressed, the overall PR looks good to me. Thanks @mroeschke let me know if you need help or assistance in addressing the changes, I can drive this if necessary.

tpcgen-cli/src/parquet.rs
```diff
@@ -52,9 +67,14 @@ where
...
     // Compute the parquet schema
-    let mut writer_properties = WriterProperties::builder()
-        .set_compression(parquet_compression)
-        .build();
+    let mut builder = WriterProperties::builder().set_compression(parquet_compression);
+    for (col, enc) in column_encodings {
+        let path = ColumnPath::from(col.as_str());
+        builder = builder
+            .set_column_encoding(path.clone(), *enc)
+            .set_column_dictionary_enabled(path, false);
```
clflushopt: if `--column-encoding` is set to `PLAIN_DICTIONARY`, `RLE_DICTIONARY` or `BIT_PACKED` the parquet write fails at runtime due to the explicit `set.column_dictionary_enabled(path, false)`.

tpcgen-cli/src/parquet.rs
```diff
@@ -52,9 +67,14 @@ where
     let schema = first_iter.schema();
 
     // Compute the parquet schema
-    let mut writer_properties = WriterProperties::builder()
-        .set_compression(parquet_compression)
-        .build();
+    let mut builder = WriterProperties::builder().set_compression(parquet_compression);
+    for (col, enc) in column_encodings {
+        let path = ColumnPath::from(col.as_str());
```
clflushopt: Since column names are never validated this will silently succeed with default encoding and the user will not notice that their override took effect.

clflushopt (changes requested):
I tested this and found a couple bugs.

tpcgen-cli/src/tpch_cli/output_plan.rs
mroeschke: Ah thanks, done in e9f072039ee65fe0d42743b68e13498449686eb2

tpcgen-cli/src/tpch_cli/output_plan.rs
mroeschke: Sure, done in 35b115f43cdd0d8c6a959a39f4c6d66d0fabb867

tpcgen-cli/src/tpch_cli/output_plan.rs
mroeschke: Sure changed in f1c66852fa334b093e66ce5c979bbafe9496eb5b (hopefully incorporates the simplifications of https://github.com/datafusion-contrib/tpcgen-rs/pull/405#discussion_r3882633455 and https://github.com/datafusion-contrib/tpcgen-rs/pull/405#discussion_r3882642674)

tpcgen-cli/src/parquet.rs
mroeschke: Sure in e3bb0eab0a8ee0e532407777e25238aa9cfee763, the CLI should raise now if a missing column name was provided

tpcgen-cli/src/parquet.rs
mroeschke: Thanks, in e3bb0eab0a8ee0e532407777e25238aa9cfee763, hopefully the error message is helpful in this case

clflushopt (approved):
Will wait for @alamb to review before merging as he is more experienced with Parquet, otherwise LGTM.

tpcgen-cli/src/tpch_cli/cli.rs
```diff
@@ -274,6 +275,23 @@ struct ParquetArgs {
     /// Typical values range from 10MB to 100MB.
     #[arg(long, default_value_t = DEFAULT_PARQUET_ROW_GROUP_BYTES)]
     row_group_bytes: i64,
+
+    /// Per-column Parquet encodings (overrides writer defaults).
+    ///
+    /// Format: `COLUMN=ENCODING[,COLUMN=ENCODING...]`
+    ///
+    /// Example: `l_comment=DELTA_LENGTH_BYTE_ARRAY,l_shipinstruct=DELTA_LENGTH_BYTE_ARRAY`
```
alamb, line 283: I tested this out

```shell
andrewlamb@Andrews-MacBook-Pro-3:~/Software/tpcgen-rs$ ./target/release/tpcgen-cli tpch parquet --column-encoding l_comment=DELTA_LENGTH_BYTE_ARRAY,l_shipinstruct=DELTA_LENGTH_BYTE_ARRAY
Error: column 'l_comment' for --column-encoding not found in table(s): nation, region, part, supplier, partsupp, customer, orders
```

Which was confusing -- I think the issue is that the program was trying to apply the encodings to each table, when the l_comment column was only applicable to `lineitem`

When I restricted to `lineitem` then it works

```shell
andrewlamb@Andrews-MacBook-Pro-3:~/Software/tpcgen-rs$ ./target/release/tpcgen-cli tpch parquet --tables=lineitem --column-encoding l_comment=DELTA_LENGTH_BYTE_ARRAY,l_shipinstruct=DELTA_LENGTH_BYTE_ARRAY
lineitem               [==================] (100%) done
```

And verified that the column was encoded

```shell
 parquet-layout lineitem.parquet
...
          "path": "l_shipinstruct",
          "has_offset_index": true,
          "has_column_index": true,
          "has_bloom_filter": false,
          "pages": [
            {
              "compression": "snappy",
              "encoding": "delta_length_byte_array",
              "page_type": "data_page_v1",
              "offset": 2787145,
              "compressed_bytes": 53002,
              "uncompressed_bytes": 254747,
              "header_bytes": 23,
              "num_values": 20096
            },
            {
              "compression": "snappy",
              "encoding": "delta_length_byte_array",
              "page_type": "data_page_v1",
              "offset": 2840170,
              "compressed_bytes": 55300,
              "uncompressed_bytes": 265377,
              "header_bytes": 23,
              "num_values": 20928
            },
...
```

tpcgen-cli/src/parquet.rs
```diff
@@ -23,6 +24,103 @@ pub trait IntoSize {
...
+        return Err(format!("expected COLUMN=ENCODING, got: '{s}'"));
+    }
+    let encoding = Encoding::from_str(encoding).map_err(|e| e.to_string())?;
+    Ok((name.to_string(), encoding))
+}
+
+/// Returns `true` if `encoding` can be used as a column's primary (non-dictionary)
+/// encoding for `physical_type`.
+///
+/// Mirrors the parquet crate's own encoder selection: the generic path
```
alamb: What is the reason to duplicate this error check in this crate? Won't the parquet crate generate the same error if we configure the writer with an incompatible set of encodings?

If we do need to do a fail fast check in this project, could you please file a ticket upstream in arrow-rs project https://github.com/apache/arrow-rs and leave a link to that ticket in the comments
The idea would be to expose this function as a real API so it will stay in sync with the parquet implementation. 

Then we can use this implementation  until it has been added upstrea,

alamb: I think remvoing all the custom error logic would also make this PR much smaller -- we could just verify that the writer errors when misconfigured and rely on the upstream arrow tests

alamb (commented):
Thank you for this PR @mroeschke  and for the reviews @clflushopt 

It looks close to me. I think we need to:
1. Make the command line argument handling better (don't error if the encoding applies to any column in one of the encoded tables). 
2. Reconsider all the encoding validation (it isn't clear why we need special encoding validation here that duplicates what is in arrow-rs)

Otherwise this looks good to me

kevinjqliu (approved):
LGTM. I think we can merge this as is and follow up with a few more improvements. 

I opened #427 to get table schemas without initializing generators. I also have another local change that will catch the incompatible encodings before generation (avoids the panic).

tpcgen-cli/src/parquet.rs
kevinjqliu: I opened https://github.com/apache/arrow-rs/issues/10964 to track this

Issue #335, opened 2026-07-06
[TPC-DS] Support specific part generation for csv, dat and parquet output


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

  tpcgen tpch has these common options that tpcgen tpcds does not:

```
  - -p, --parts <PARTS>
  - --part <PART>
```

<details><summary>Output of <pre>tpcgen tpch</pre> </summary>
<p>

```shell
TPC-H data

Usage: tpcgen-cli tpch [OPTIONS]
       tpcgen-cli tpch <COMMAND>

Commands:
  tbl      Generate TBL (pipe-delimited) output
  csv      Generate CSV output with CSV-specific options
  parquet  Generate Apache Parquet output with Parquet-specific options
  help     Print this message or the help of the given subcommand(s)

Options:
  -s, --scale-factor <SCALE_FACTOR>
          Scale factor to create

          [default: 1]

  -o, --output-dir <OUTPUT_DIR>
          Output directory for generated files (default: current directory)

          [default: .]

  -T, --tables <TABLES>
          Which tables to generate (default: all)

          Possible values:
          - region:   Region table (alias: r)
          - nation:   Nation table (alias: n)
          - supplier: Supplier table (alias: s)
          - customer: Customer table (alias: c)
          - part:     Part table (alias: P)
          - partsupp: PartSupp table (alias: S)
          - orders:   Orders table (alias: O)
          - lineitem: LineItem table (alias: L)

  -p, --parts <PARTS>
          Number of part(itions) to generate. If not specified creates a single file per table

      --part <PART>
          Which part(ition) to generate (1-based). If not specified, generates all parts

  -n, --num-threads <NUM_THREADS>
          The number of threads for parallel generation, defaults to the number of CPUs

          [default: 16]

  -v, --verbose
          Verbose output

          When specified, sets the log level to `info` and ignores the `RUST_LOG` environment variable. When not specified, uses `RUST_LOG`

  -q, --quiet
          Quiet mode - only show error-level logs

      --stdout
          Write the output to stdout instead of a file

      --no-progress
          Disable progress bars during data generation.

          Bars are also auto-suppressed by `--quiet`, `--stdout`, or when stderr is not a terminal.

  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version
```

</p>
</details>

alamb:
[...]

PR #406, opened 2026-08-28
feat: support generating specific parts in tpcds

clflushopt:
Addresses #335 by adding support for `-p/--parts` and `--part` to `tpcgen tpcds`, matching the flags `tpcgen tpch` already has (#218), this follows the approach @alamb audited in dsdgen's `-PARALLEL`/`-CHILD` options and described extensively in #335.

The actual changes are mostly gluing together the existing mechanics that we implemented from the port we didn't introduce a lot of custom logic :

- Port dsdgen's `split_work` (`tools/parallel.c`) into `Session`: tables under 1M source rows are generated entirely by chunk 1 (dsdgen's small-table rule), larger tables split evenly with the remainder spread over the first chunks.
- Wiring the existing `RandomNumberStream::skip_rows` fast-forward into the DAT/CSV row-generator driver via `skip_rows_until_starting_row_number`, so each chunk starts at the right row deterministically.
- Parquet's per-table row-group plan (`TpcdsGenerationPlan`) now takes an explicit row range, so `--parts` shrinks it the same way.
- File naming follows `tpchgen-cli`'s convention: `<table>/<table>.<part>.<ext>`.
- Progress bar totals are scaled to each chunk's row range instead of the whole table.

I also added new unit tests and validated SF=10 by concatenating a 4 way split of `store_sales` and checked the progress bar TUI rendering.


tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -1000,3 +1000,262 @@ fn test_tpcgen_cli_tpcds_help_lists_tables() {
...
+        .arg("--parts")
+        .arg("4")
+        .assert()
+        .success();
+
+    let chunk_one =
+        fs::read_to_string(temp_dir.path().join("reason/reason.1.dat")).expect("chunk 1 exists");
+    assert_eq!(chunk_one.lines().count(), 35, "chunk 1 has every row");
+
+    for chunk in 2..=4 {
```
kevinjqliu: why create empty chunks / files rather than just skipping?

tpcdsgen/src/config/session.rs
```diff
@@ -1,5 +1,55 @@
 use crate::config::{CompatMode, Scaling, Table};
 use crate::error::{InvalidOptionError, Result};
+use std::ops::RangeInclusive;
+
+/// Tables with fewer source rows than this are not split across chunks:
+/// chunk 1 generates the whole table and every other chunk generates none of
+/// it. Matches dsdgen's `tools/parallel.c` (`kRowsetSize` is only computed
```
alamb, line 7: I recommend adding a link to `tools/parallel.c`'s source

Also the details about ` (kRowsetSize is only computed ...`) seems like an implementation detail of the C code that is not relevant here

tpcdsgen/src/config/session.rs
```diff
@@ -1,5 +1,55 @@
...
+/// when `nTotalRows` is at least this many rows).
+const SMALL_TABLE_ROW_THRESHOLD: i64 = 1_000_000;
+
+/// Split `total_rows` into `total_chunks` pieces and return the 1-based
+/// `(first_row, row_count)` for `chunk_number`.
+///
+/// Ports dsdgen's `split_work` (`tools/parallel.c`) / Trino's
```
alamb, line 14: again, here I think it would be most helpful to put a link to the relevant code and not try to summarize how it works (basically "Ports dsdgen's `split_work`" is probably enough)

tpcdsgen/src/config/session.rs
```diff
@@ -82,6 +136,25 @@ impl Session {
...
+    /// Return the total number of chunks this session's table generation is
+    /// split across.
+    pub fn get_total_chunks(&self) -> i32 {
+        self.total_chunks
+    }
+
+    /// Return the 1-based, inclusive range of `table`'s source rows this
+    /// session's chunk is responsible for generating.
+    ///
+    /// Uses [`Table::source_table`] so a returns table (e.g.
```
alamb, line 148: I found the "uses ..." part  confusing -- it seems like the point is that this method works for all types of tables. I found the description of the internal implementation confusing

tpcgen-cli/src/tpcds_cli/plan.rs
```diff
@@ -26,29 +26,44 @@ pub(super) struct TpcdsGenerationPlan {
...
-    pub(super) fn new(table: Table, scaling: &Scaling, row_group_bytes: usize) -> Self {
-        let source_rows = scaling.get_row_count(table.source_table());
-        let estimated_bytes = source_rows.saturating_mul(estimated_bytes_per_source_row(table));
+    /// `row_group_bytes`, restricted to `row_range` of the table's source
+    /// rows.
+    ///
+    /// `row_range` is typically a whole table (`1..=source_rows`, see
+    /// [`Self::new`]) or one `--parts`/`--part` chunk (see
+    /// [`tpcdsgen::config::Session::get_source_row_range`]); either way the
+    /// row groups it produces cover exactly `row_range`, so the row group
```
alamb, line 35: "so the row group
    /// count naturally shrinks for a smaller chunk instead of needing a
    /// separate scaling step." 

seems somewhat redundant -- just saying it covers exactly `row_range` seems like enough

tpcgen-cli/src/tpcds_cli/progress.rs
```diff
@@ -22,14 +23,24 @@ pub(super) enum TableProgress {
     },
 }
 
+/// Number of rows in an inclusive row range, clamped to non-negative (an
+/// empty range, e.g. from a chunk that got no rows under the small-table
```
alamb, line 27: What is the reference to the small table rule?  It seems like we could probably drop that and just document this function

tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -1000,3 +1000,262 @@ fn test_tpcgen_cli_tpcds_help_lists_tables() {
...
+        concatenated.extend(fs::read(&path).unwrap_or_else(|err| panic!("{path:?} exists: {err}")));
+    }
+
+    assert_eq!(
+        concatenated, unsplit,
+        "Expected concatenated --parts chunks to reproduce the unsplit DAT output"
+    );
+}
+
+/// Test the same concatenation property for CSV, ignoring each part's own
```
alamb, line 1135: 👍

alamb (commented):
Thanks @clflushopt  -- this looks great. I left some comment cleanup suggestions, but also I think we need to add more end to end coverage (see inline coments) before we merge this

Some minor details

## Strange output  for mulit-part tabled

When I ran this:
```shell
./target/release/tpcgen-cli tpcds --parts=10
```
Each part seems to have gotten its own line (but the name was the same)
Image

Vs running tpch there is just one entry per file even when creating multiple parts
```
andrewlamb@Andrews-MacBook-Pro-3:~/Software/tpcgen-rs$ ./target/release/tpcgen-cli tpch --parts=10
nation                 [==================] (100%) done                                                                                                                                                            region                 [==================] (100%) done                                                                                                                                                            part                   [==================] (100%) done                                                                                                                                                            supplier               [==================] (100%) done                                                                                                                                                            partsupp               [==================] (100%) done                                                                                                                                                            customer               [==================] (100%) done                                                                                                                                                            orders                 [==================] (100%) done                                                                                                                                                            lineitem               [==================] (100%) done
```

## `--parts 1` diverges from tpch: 

in tpch, parts=1 generates 

```shell
/target/release/tpcgen-cli tpch --parts=1
```

Makes  directories with a single file
```shell
andrewlamb@Andrews-MacBook-Pro-3:~/Software/tpcgen-rs$ ls lineitem/
lineitem.1.tbl
```

But tpcds with a single part makes single files
```shell
./target/release/tpcgen-cli tpcds --parts=1
```

Makes single files like `warehouse.dat`

```shell
ls *.dat
...
-rw-r--r--@  1 andrewlamb  staff   585B Aug 30 07:15 warehouse.dat
-rw-r--r--@  1 andrewlamb  staff   5.6K Aug 30 07:15 web_page.dat
-rw-r--r--@  1 andrewlamb  staff   9.4M Aug 30 07:15 web_returns.dat
-rw-r--r--@  1 andrewlamb  staff   140M Aug 30 07:15 web_sales.dat
-rw-r--r--@  1 andrewlamb  staff   8.6K Aug 30 07:15 web_site.dat
...
```

It would be nice to have the same behavior (tpcds seems better to me actually)

tpcgen-cli/src/tpcds_cli/progress.rs
```diff
@@ -42,20 +53,47 @@ pub(super) fn register_table(
...
+    // row), so scale that full-table estimate down by the same fraction of
+    // source rows this chunk covers.
+    let chunk_returns_row_count = |sales_table: Table, returns_table: Table| {
+        let full_source = session.get_scaling().get_row_count(sales_table);
+        if full_source <= 0 {
+            return 0;
+        }
+        let full_returns = session.get_scaling().get_row_count(returns_table);
+        let chunk_source = chunk_row_count(sales_table);
+        (full_returns * chunk_source) / full_source
```
kevinjqliu, line 71:
```suggestion
        ((i128::from(full_returns) * i128::from(chunk_source))
                / i128::from(full_source)) as i64
```

`full_returns` and `chunk_source` are both `i64`. This can overflow with valid inputs. For example, SF 100,000 with `--parts 4` has `full_returns = 16B` and `chunk_source ≈ 4B`, producing a `64e18` intermediate value before division, above `i64::MAX`. Using `i128` for intermediate value here is safer

tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -1000,3 +1000,262 @@ fn test_tpcgen_cli_tpcds_help_lists_tables() {
...
+
+    for chunk in 2..=4 {
+        let path = temp_dir.path().join(format!("reason/reason.{chunk}.dat"));
+        let contents = fs::read_to_string(&path)
+            .unwrap_or_else(|err| panic!("Expected {path:?} to exist: {err}"));
+        assert!(
+            contents.is_empty(),
+            "Expected chunk {chunk} of `reason` to be empty, got: {contents:?}"
+        );
+    }
```
kevinjqliu, line 1083: nit: i find it a little awkward that we end up with a bunch of empty files, but this is consistent with the current tpchgen behavior. 
We might want to revisit this in the future.

tpcgen-cli/src/tpcds_cli/generate.rs
```diff
@@ -9,12 +9,36 @@
...
+        let dir = output_dir.join(table.get_name());
+        std::fs::create_dir_all(&dir)?;
+        Ok(dir.join(format!(
+            "{}.{}.{ext}",
+            table.get_name(),
+            session.get_chunk_number()
+        )))
+    } else {
+        Ok(output_dir.join(format!("{}.{ext}", table.get_name())))
+    }
```
kevinjqliu, line 39: as alamb also mentioned, this diverges from the current tpchgen logic for `--parts 1`, which produce `reason/reason.1.dat` and not `reason.dat`

tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -236,37 +244,45 @@ impl CommonArgs {
 
         std::fs::create_dir_all(&self.output_dir)?;
 
+        let parts = self.part_list()?;
+
```
kevinjqliu, line 248:
```suggestion

        let parts = self.part_list()?;

        std::fs::create_dir_all(&self.output_dir)?;

```

swap this so `self.part_list()` can fail for invalid configuration (i.e. `--parts 3 --part 4`) without creating folders

kevinjqliu (commented):
One more step closer to parity with tpchgen :) I left a couple of comments. 

I also think we need to rework the progress bar rendering in this PR. The PR creates one progress bar per table per part, so `--parts 4` produces around 100 bars. `tpchgen` uses one progress bar per table regardless of the number of parts, and I think TPC-DS should follow the same approach. The tradeoff is that we’ll show aggregate progress rather than progress for each individual part.

tpcgen-cli/src/tpcds_cli/progress.rs
kevinjqliu: Also found a similar issue and fixed it in #419

alamb:
I am thinking I may spend some time helping this PR over the line in the next day or two (I need a fun project this weekend). I'll just push directly to this branch, unless you would like me to do anything different

kevinjqliu:
happy to review. I think we're very close to getting the release out 🚀

tpcgen-cli/tests/cli_integration/tpcds.rs
alamb: I actually tried this out locally, and it turns out dsgen doesn't actually make empty files, it simply doesn't make them at all. So making part 2 of 4 for `reason` makes no file at all

I will fix this

```shell
andrewlamb@Andrews-MacBook-Pro-3:~/Software/DSGen-software-code-4.0.0/DSGen-software-code-4.0.0/tools$ rm reason_1_4.dat
andrewlamb@Andrews-MacBook-Pro-3:~/Software/DSGen-software-code-4.0.0/DSGen-software-code-4.0.0/tools$ rm *.dat
rm: *.dat: No such file or directory
andrewlamb@Andrews-MacBook-Pro-3:~/Software/DSGen-software-code-4.0.0/DSGen-software-code-4.0.0/tools$ ls *.dat
ls: *.dat: No such file or directory
andrewlamb@Andrews-MacBook-Pro-3:~/Software/DSGen-software-code-4.0.0/DSGen-software-code-4.0.0/tools$ ./dsdgen -PARALLEL 4 -CHILD 1 -TABLE reason -SCALE 1
dsdgen Population Generator (Version 4.0.0)
Copyright Transaction Processing Performance Council (TPC) 2001 - 2021
Warning: This scale factor is valid for QUALIFICATION ONLY
andrewlamb@Andrews-MacBook-Pro-3:~/Software/DSGen-software-code-4.0.0/DSGen-software-code-4.0.0/tools$ ls *.dat
reason_1_4.dat
andrewlamb@Andrews-MacBook-Pro-3:~/Software/DSGen-software-code-4.0.0/DSGen-software-code-4.0.0/tools$ ./dsdgen -PARALLEL 4 -CHILD 2 -TABLE reason -SCALE 1
dsdgen Population Generator (Version 4.0.0)
Copyright Transaction Processing Performance Council (TPC) 2001 - 2021
Warning: This scale factor is valid for QUALIFICATION ONLY
andrewlamb@Andrews-MacBook-Pro-3:~/Software/DSGen-software-code-4.0.0/DSGen-software-code-4.0.0/tools$ ls *.dat
reason_1_4.dat
```

tpcdsgen/src/config/session.rs
```diff
@@ -1,5 +1,57 @@
...
+use std::ops::RangeInclusive;
+
+/// Threshold above which tables are not split across chunks. For tables with
+/// fewer than this many rows, chunk 1 generates the whole table . Matches the
+/// C `dsdgen`'s `kTotalRows < 1000000` check in `tools/parallel.c` [1] and
+/// Trino's `Parallel.SMALL_TABLE_THRESHOLD` [2].
+///
+/// [1]: https://github.com/gregrahn/tpcds-kit/blob/5a3a81796992b725c2a8b216767e142609966752/tools/parallel.c#L75
+/// [2]: https://github.com/trinodb/tpcds/blob/b594136818cc95bd6b34a352327611b329017281/src/main/java/io/trino/tpcds/Parallel.java#L28
+const SMALL_TABLE_ROW_THRESHOLD: u64 = 1_000_000;
```
alamb, line 12: the only thing that I think we should reconsider is this split logic -- this mirrors how the dsdgen program does it, but it seems somewhat aributrary to me and is not consistent with how tpcgen-cli does it

So specifically we could consider splitting files with fewer than 1M rows, but I also think this is fine too

tpcdsgen/src/config/session.rs
```diff
@@ -315,4 +444,401 @@ mod tests {
         let session = Session::default();
         session.get_only_table_to_generate();
     }
+
+    #[test]
```
alamb, line 448: This looks like a lot of tests but I think it covers the right logic to encode in tests -- specifically the exact sizes of the tables across different scale factors. It is mostly just a bunch of data encoded to ensure we catch behavior change.

There are a few integration tests that do this with actual data below, however, they take a while so I wanted size coverage without the end to end test

tpcgen-cli/src/tpcds_cli/parquet.rs
```diff
@@ -139,15 +145,44 @@ impl Parquet {
         }
 
         // Plan each table and pre-register the row group totals so trackers
-        // can size their bars before the first increment
-        let mut work: Vec<(Table, Session, TpcdsGenerationPlan, ProgressHandle)> = table_sessions
+        // can size their bars before the first increment.
+        let planned: Vec<(Table, Session, TpcdsGenerationPlan)> = table_sessions
```
alamb: this progress tracking stuff looks complicated -- I didn't change it, and it seems to work.

tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -501,7 +501,7 @@ fn test_tpcgen_cli_tpcds_dat_multiple_table_selection_command_forms() {
         cargo_bin_cmd!("tpcgen-cli")
             .args(*form)
             .arg("--scale-factor")
-            .arg("0")
+            .arg("0.001")
```
alamb: I changd the code to not write files with 0 length, so I had to increase the scale factor to something small in these tests

tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -1147,3 +1147,405 @@ fn test_tpcgen_cli_tpcds_help_lists_tables() {
         );
     }
 }
+
+/// Test that `--part` without `--parts` is rejected with the expected message.
```
alamb, line 1151: These are the key integration tests -- I only do a few spot checks to keep CI time reasonable, but I think it is enough to verify that the plumbing is all hooked up correctly

tpcgen-cli/src/tpcds_cli.rs
alamb: Fixed in 68dd985

tpcdsgen/src/config/session.rs
alamb: Update is I played around with this idea locally by making this parameter configurable, and it turns out to be trick to get the row number accounting to work right with the tables that have chained generators (and when the parts are concatenated they don't form the entire original table)

So I think we should leave it here for now

tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -1147,3 +1147,470 @@ fn test_tpcgen_cli_tpcds_help_lists_tables() {
...
+// ----------------
+// Test that concatenating a file created with `--parts`
+// exactly reproduces a single-file output
+//
+// Most of the tables are below the 1M minimum size, so only testing the larger
+// tables is sufficient to verify the behavior.
+//
+// For now, to avoid long test times, we test only a subset of the larger tables.
+// testing the whole thing is tracked by
+// https://github.com/datafusion-contrib/tpcgen-rs/issues/457
```
alamb, line 1330: It would be nice to test this more, but i think it takes too long today. Once I sort out parallel generation, I'll follow up with some more extensive tests
- https://github.com/datafusion-contrib/tpcgen-rs/issues/457

alamb (commented):
Alright, I think this is now ready to review again. cc @clflushopt and @kevinjqliu let me know what you think

alamb (approved):

alamb:
Sorry for the size but this PR is largely tests

tpcgen-cli/src/tpcds_cli/parquet.rs
kevinjqliu: i have one suggestion that can simplify the logic here. in this commit: https://github.com/kevinjqliu/tpchgen-rs/commit/a66c3085df19b230788142f82f3da03644cb9d33#r201136343

we can also pass the (table, sessions) pre-grouped from `tpcgen-cli/src/tpcds_cli.rs`

kevinjqliu (approved):
LGTM! I validated the changes to introduce `parts` (`total_chunks`) and `part` (`chunk_number`) to the generation logic and progress tracker.

I tested the progress ui locally. I suspect we can simplify this code a lot when we disaggregate the sales/returns table generation.

I have 1 suggestion for the progress bar logic in parquet.rs

alamb:
> I have 1 suggestion for the progress bar logic in parquet.rs

Thank you -- done. 

I will now merge this in and try and work on parallel part generation tomorrow morning. I have been nerd sniped!

alamb:
🚀

clflushopt:
thanks @alamb and @kevinjqliu sorry i couldn't finish this got busy last week PR looks in good shape

alamb:
No worries! and no problems! I am sorry the initial PR review was so delayed.

In september I am trying to clean out a bunch of my backlog projects and getting tpcgen support with TPCDS out is on that list. I think we are very close

PR #413, opened 2026-08-28
Publish crates.io packages as a workspace

kevinjqliu:
## Summary

Replace the per-crate publishing matrix with `cargo publish --workspace`.

## Why

- Cargo determines package and dependency order from workspace manifests.
- Future publishable workspace members require no workflow changes.
- Matches our lockstep release process: every artifact is version-bumped before publishing.

## Dry run

1. Temporarily bumped TPC-H artifacts from `3.0.0` to `3.0.1` and TPC-DS/unified artifacts from `0.1.0` to `0.1.1`.
2. Updated all internal dependency constraints to the temporary versions.
3. Ran `cargo publish --workspace --dry-run --allow-dirty`.
4. Confirmed all six crates packaged and compiled, with dependencies ordered before dependents.
5. Restored the temporary version changes.

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -13,26 +13,13 @@ jobs:
     permissions:
       contents: read
       id-token: write     # Required for OIDC token exchange
-    strategy:
-      # Publish package one by one instead of flooding the registry
-      max-parallel: 1
-      matrix:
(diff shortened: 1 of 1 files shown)


alamb (approved):
Let's give it a try! I don't really know how to test this but I suppose we can fix it if needed  if it breaks

PR #415, opened 2026-08-28
ci: audit Rust dependencies

kevinjqliu:
## Summary

Add a `cargo audit` workflow to catch vulnerable dependency updates and newly published advisories. It runs for pull requests, pushes to `main`, weekly schedules, and manual dispatches using `cargo-audit` 0.22.2 with SHA-pinned actions, read-only permissions, and a ten-minute timeout.

The workflow audits `Cargo.lock` when present; otherwise, `cargo audit` generates the dependency resolution.

## Validation

- `actionlint .github/workflows/audit.yml`
- `cargo audit` scanned 200 dependencies with no vulnerabilities

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,30 @@
+name: Dependency Audit
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+  schedule:
+    - cron: "17 6 * * 1"
+  workflow_dispatch:
+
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thanks @kevinjqliu 

I also double checked and we are already setup with Dependabot: https://github.com/datafusion-contrib/tpcgen-rs/blob/main/.github/dependabot.yml

Issue #423, opened 2026-09-06
[Epic] Align TPC-H and TPC-DS CLI flags and behavior for v4


kevinjqliu:
TPC-H was implemented first. Most of its flags apply just as well to TPC-DS: output directories, stdout, partitions, threads, and logging should work the same way in both.

This issue tracks that alignment for `tpcgen-cli tpch` and `tpcgen-cli tpcds`. The legacy `tpchgen-cli` command should stay consistent with `tpcgen-cli tpch`.

## What should match

Unless marked CSV or Parquet, generation flags should apply to every format, including the default TBL/DAT command. Help and version flags should match at corresponding command levels. Some already match; the checklist below tracks the remaining work.

| Flag | Expected behavior in both benchmarks |
|---|---|
| `-s`, `--scale-factor` | Default `1`; accept `0` through `100000`, inclusive; reject invalid values. |
| `-o`, `--output-dir` | Default `.`; create directories when needed. Skip existing final files by default, with a warning unless quiet. Use `--overwrite` to allow replacement. Never publish a partial final file. |
| `--stdout` | Write data to stdout, diagnostics to stderr. Disable progress bars and do not create output files or directories. Define consistent supported combinations of stdout, table selection, and partition selection for both benchmarks. |
| `-T`, `--tables` | Default to all tables. Accept comma-separated selections, remove duplicates, and write only selected tables. Accept full names case-insensitively; show the invalid value and valid choices on errors. |
| `-p`, `--parts` | Partition output using matching validation and file-naming conventions, while preserving benchmark-specific exceptions. TPC-H does not split `nation` or `region`. |
| `--part` | Select one 1-based partition; otherwise generate all partitions. Match the existing TPC-H behavior with `--parts`. |
| `-n`, `--num-threads` | Accept positive counts, default to CPU count, and support parallel generation in every format. |
| `-v`, `--verbose` | Enable info-level logs, override `RUST_LOG`, and conflict with `--quiet`. Report the table, destination, and successful completion. |
| `-q`, `--quiet` | Show errors only. Suppress progress bars and existing-file warnings; conflict with `--verbose`. |
| `--no-progress` | Disable progress bars. Also suppress them automatically for quiet mode, stdout, and non-terminal stderr. |
| `--delimiter` (CSV) | Default comma. Accept the same delimiters and escape sequences. Every accepted delimiter must produce valid headers and data rows. |
[...]

PR #424, opened 2026-09-06
fix: align TPC-H Parquet row-group validation with TPC-DS

kevinjqliu:
Aligns TPC-H with the TPC-DS argument validation added in #349, addressing the row-group size checklist item in #423.

TPC-H previously panicked on zero and could produce an invalid empty file for negative `--row-group-bytes` values. Both CLIs now share an `i64` parser that rejects non-positive values before generation with `must be greater than zero`.

TPC-DS also uses `i64` throughout row-group planning, removing its unchecked `usize`-to-`i64` cast. Both commands now accept positive values up to `i64::MAX`; defaults are unchanged. The TPC-H fix also covers standalone `tpchgen-cli parquet`.

Both benchmarks use concise option descriptions (help excerpts):

```text
$ tpcgen-cli tpch parquet --help
      --row-group-bytes <ROW_GROUP_BYTES>
          Target row-group size in bytes

$ tpcgen-cli tpcds parquet --help
      --row-group-bytes <ROW_GROUP_BYTES>
          Target row-group size in bytes
```

### DIFF
--- a/tpcgen-cli/src/tpcds_cli.rs
+++ b/tpcgen-cli/src/tpcds_cli.rs
@@ -25,8 +25,6 @@ mod progress;
 
 type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
-const DEFAULT_TPCDS_PARQUET_ROW_GROUP_BYTES: usize = DEFAULT_PARQUET_ROW_GROUP_BYTES as usize;
-
 enum OutputFormat {
     Dat(dat::Dat),
     Csv(csv::Csv),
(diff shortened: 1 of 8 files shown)


tpcgen-cli/tests/cli_integration.rs
kevinjqliu, line 1: this tests both `tpch` and `tpcds` subcommands

tpcgen-cli/src/tpcds_cli/parquet.rs
```diff
@@ -25,15 +25,15 @@ use tpcdsgen_arrow::{
 pub(super) struct Parquet {
     output_dir: PathBuf,
     compression: Compression,
-    row_group_bytes: usize,
+    row_group_bytes: i64,
```
kevinjqliu, line 28: `usize` -> `i64` refactor to align with tpch

tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -464,15 +463,6 @@ fn parse_delimiter(s: &str) -> std::result::Result<char, String> {
     Ok(parsed)
 }
 
-fn parse_row_group_bytes(s: &str) -> std::result::Result<usize, String> {
```
kevinjqliu, line 467: moved to shared `args.rs`

alamb (approved):
makes sense to me -- thank you @kevinjqliu

PR #433, opened 2026-09-07
ci: run full packaging in the merge queue

kevinjqliu:
Python packaging is expensive and does not need to run on every pull request or again after the merge queue succeeds.

This PR change runs the full wheel matrix and both CLIs' source distributions in the **merge queue only**. 
- Pull requests and post-merge pushes to `main` skip packaging. 
- Release workflows are unchanged and continue to build and upload all packages.


alamb:
What would you thik about just running the packaging jobs when they were added to the merge queue (rather than adding new tests?)

alamb (approved):
makes sense to me -- thanks @kevinjqliu

kevinjqliu:
> What would you thik about just running the packaging jobs when they were added to the merge queue (rather than adding new tests?)

yea thats a good idea. Originally I also wanted to do a smoke test (w/ a few select OS) for PR, but i think it makes sense to just move everything to the merge queu

alamb:
🚀

PR #434, opened 2026-09-07
ci: remove lint dependencies and add Rust caching

kevinjqliu:
This PR makes two changes to reduce CI waiting and repeated compilation:

1. **Remove `needs: lint` so independent jobs run in parallel.** Doc tests, workspace tests, documentation builds, and packaging previously waited for lint to finish. They can now start alongside it. The required `Rust checks` status still requires every job to pass.
2. **Cache Rust dependencies to reuse previous build work.** Add `Swatinem/rust-cache` to the Rust check jobs and replace `actions/cache` in conformance and benchmarks. This reuses downloaded dependencies and their compiled artifacts, with Rust-aware cache keys and a consistent write policy: only `main` saves caches; other refs can restore them.

Test commands and packaging coverage are unchanged.

### DIFF
--- a/.github/workflows/build-py-packages.yml
+++ b/.github/workflows/build-py-packages.yml
@@ -90,6 +90,7 @@ jobs:
           args: --locked --release --out dist
           working-directory: tpchgen-cli
           manylinux: ${{ matrix.manylinux || '' }}
+          sccache: true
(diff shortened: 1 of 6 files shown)


alamb (approved):
Looks good to me. Thanks @kevinjqliu

PR #442, opened 2026-09-09
Align TPC-DS canonical column names

### LINKED ISSUE #437: [TPC-DS] Schema mismatch vs TPC spec
Some TPCDS column names do not match the TPC spec: https://www.tpc.org/TPC_Documents_Current_Versions/pdf/[TPC-DS_v4.0.0.pdf](https://www.tpc.org/TPC_Documents_Current_Versions/pdf/TPC-DS_v4.0.0.pdf)

The following deviate:
  - `cr_return_amount_inc_tax` instead of `cr_return_amt_inc_tax`
  - `ib_income_band_id` instead of `ib_income_band_sk`
  - `r_reason_description` instead of `r_reason_desc`
  - `s_tax_precentage` instead `s_tax_percentage` correction
  - `wr_store_credit` instead of `wr_account_credit`

Java/C implementations had these typos, wouldn't it be better to adhere to the official TPC spec?

In [LakeBench](https://github.com/microsoft/LakeBench) we made a strategic decision to adhere to spec so I'm currently evaluating whether we should intentionally adopt the typos just so that we can leverage this amazing project. We previously did not as the DuckDB generator did follow the spec without typos and this is what we elected to use for data generation.

mwc360:
## Summary
Aligns TPC-DS output column names with the canonical specification without changing data types or generated values.
 
Closes #437 

 ## Changes 
- Corrects four canonical column names across Arrow schemas and CSV headers:
     - `cr_return_amount_inc_tax` → `cr_return_amt_inc_tax`
     - `ib_income_band_id` → `ib_income_band_sk`
     - `r_reason_description` → `r_reason_desc`
     - `wr_store_credit` → `wr_account_credit`
- Renames related internal row fields, generator locals, Arrow buffers, CLI examples, and column-encoding tests.
- Preserves deprecated compatibility getters for `get_ib_income_band_id()` and `get_r_reason_description()`.
- Adds focused Arrow, CSV, and CLI regression coverage for the canonical names and removal of obsolete output names.
 
 ## Scope
This PR intentionally does not change decimal precision, integer widths, nullability, address types, or generated data values.

`s_tax_precentage` is also left as is: the spec PDF spells it `s_tax_percentage`, but the dsdgen toolkit DDL and Trino both use `s_tax_precentage` (misspelled), and this repo follows the toolkit.
 
 ## Validation
- `cargo fmt --all -- --check`
- `cargo test -p tpcdsgen`
- `cargo test -p tpcdsgen-arrow --test schema`
- Targeted TPC-DS CSV and CLI column-encoding tests


kevinjqliu (approved):
Thanks for the PR! Verified all 25 tables match the toolkit DDL and the v3.2.0 spec PDF now (only `s_tax_precentage` remains, which the toolkit itself misspells).

I have a few nit changes on the tests, pushed here so you can cherry-pick if you agree:
https://github.com/kevinjqliu/tpchgen-rs/commit/79c69121221fc6c8fe3dae0e564938f9820c3eae

- `tpcdsgen-arrow/tests/schema.rs` now checks every table's full column list against the canonical names, so it catches the next typo or reorder anywhere, not just these 4. The "obsolete name is absent" checks are gone since they can't fail once the canonical name is present.
- Dropped `test_tpcgen_cli_tpcds_parquet_rejects_obsolete_column_name` and `test_tpcgen_cli_tpcds_csv_uses_canonical_column_names`. The first is the existing typo test with a different string; the CLI doesn't know about old names, it's just an unknown column. The second generates 6 CSV files to check 4 `&'static str` constants, and `reparse.rs` already validates CSV headers against the Arrow schemas, so the full schema test covers headers for free.
- `validate_column_encodings_rejects_dictionary_encoding` now asserts on the error text. It only checked `is_err()`, and the encoding is rejected before the column is looked up, so it passed for any column string.

kevinjqliu (approved):
LGTM!

PR #443, opened 2026-09-09
Align TPC-DS canonical data types

### LINKED ISSUE #282: [TPC-DS] Research:  unconditiional use of `Decimal128(38, 2)` in Arrows

nit: should we always use `Decimal128(38, 2)`?

looks like some columns have specific decimal precision and scale 
https://github.com/databricks/tpcds-kit/blob/1b7fb7529edae091684201fab142d956d6afd881/tools/tpcds.sql#L57

https://www.tpc.org/tpc_documents_current_versions/pdf/tpc-ds_v1.4.0.pdf (grep for "decimal")

_Originally posted by @kevinjqliu in https://github.com/clflushopt/tpchgen-rs/pull/276#discussion_r3487362296_

mwc360:
## Summary
 
Align TPC-DS Arrow and Parquet data types with canonical TPC-DS DDL, while retaining 64-bit order/ticket identifiers needed at large scale factors. Recalibrate the CLI's Parquet row-size estimates for the updated schema.
 
Column-name alignment is intentionally excluded and handled in a separate PR which should be merged first (#442)

Closes #282 
 
## Schema changes
 
- Narrow 124 integer fields from `Int64` to `Int32`, bringing the schema to 183 `Int32` fields and six `Int64` order/ticket fields.
- Retain `Int64` for sales and returns order/ticket identifiers, whose domains can exceed the signed 32-bit range at SF100000.
- Replace generic `Decimal128(38,2)` declarations with the applicable precision and scale as defined in the TPCDS spec:
 
| Columns | Arrow decimal type |
|---|---|
| GMT offsets and tax percentages | `Decimal128(5,2)` |
| Prices and sales/returns monetary values | `Decimal128(7,2)` |
| `promotion.p_cost` | `Decimal128(15,2)` |
 
- Scale whole-hour GMT offsets correctly when constructing decimal arrays, without changing their DAT/CSV representation.
- Emit address street numbers as strings rather than integers.
- Align field nullability with canonical `NOT NULL` constraints.
 
## Conversion safety and compatibility
 
Generator arithmetic, row models, and public Rust APIs remain unchanged. Integer narrowing occurs at the Arrow boundary using checked conversions, preserving null/sentinel handling and preventing silent overflow.
 
The data types align with what is canonical in the TPCDS 4.0.0 spec. The spec technically does not specify the integer size but existing data generators commonly map most identity columns to Int32 with order/ticket identifiers being Int64 as an exception. This results in faster parquet scanning and less uncompressed data on disc.
 
## Parquet row-group planning
 
Refresh the static per-table uncompressed byte estimates using offline SF100 measurements. Large-table samples were approximately 128 MiB uncompressed when using a 128MB `--row-group-bytes` target. Fractional estimates avoid disproportionate rounding errors for narrow tables such as Inventory.
 
Also use proper ceiling division when calculating group counts, avoiding an unnecessary extra group when estimated bytes are an exact multiple of the target.
 
## Coverage
 
Add full-table schema conformance coverage, checked-conversion boundary cases, SF100000 key-domain checks, and persisted Parquet schema assertions. Update planner and integration expectations for the new estimates, retaining source-range equivalence and thread-count reproducibility coverage.

## Benchmark
| Metric | `main` | This PR | Change |
|---|---:|---:|---:|
| Median generation time | 1,408.43 s | 1,262.36 s | **10.4% less time** |
| ZSTD(1) Parquet file size | 25.045 GiB | 24.663 GiB | **1.53% smaller** |
| Uncompressed encoded Parquet size | 118.291 GiB | 31.873 GiB | **73.1% smaller** |
| Row groups | 954 | 274 | |
| Generated rows | 959,037,906 | 959,037,906 | Unchanged |
 
### Methodology
 
- Release builds with identical locked dependencies and Rust 1.97.1.
- AMD Ryzen 9 7950X workstation running Windows 11 Enterprise.
- ZSTD(1), 128 MiB uncompressed row-group target, 8 generation threads, default column encodings, and no external partitioning.
- One SF0.01 warmup per revision, followed by sequential SF100 runs in main / PR / PR / main / main / PR order.
- Timings include generation, encoding, and file writes; exclude compilation, metadata inspection, and cleanup. OS caches were not flushed.
 
| Run | `main` | This PR |
|---|---:|---:|
| 1 | 1,408.43 s | 1,282.12 s |
| 2 | 1,425.78 s | 1,262.36 s |
| 3 | 1,365.75 s | 1,113.15 s |


tpcgen-cli/src/tpcds_cli/plan.rs
```diff
@@ -75,91 +77,45 @@ impl IntoIterator for TpcdsGenerationPlan {
...
 /// is also what the previous `ArrowWriter` based implementation limited.
 ///
-/// Measured from files generated at scale factor 1: the total uncompressed
-/// bytes, computed using datafusion-cli:
+/// Measured offline at scale factor 100 using the default column encodings.
+/// Large tables were sampled in single row groups near 128 MiB uncompressed
+/// (119-165 MiB); smaller tables were measured in full. Sizes are approximate:
+/// cardinality, row-group size, and column encodings affect encoding efficiency.
 ///
-/// You can verify these numbers using
```
clflushopt, line 81: non-issue: not sure how useful this comment was but vis-a-vis your changes i think its okay to remove

clflushopt (approved):
lgtm thanks @mwc360

tpcgen-cli/src/tpcds_cli/plan.rs
alamb: I personally would request we keep it as a cookbook for how to remeasure the sizes

tpcdsgen-arrow/tests/schema.rs
```diff
@@ -593,6 +704,303 @@ fn canonical_schemas(session: &Session) -> Vec<(&'static str, SchemaRef, &'stati
     ]
 }
 
+#[test]
+fn schemas_match_expected_columns_and_canonical_types() {
```
alamb, line 681: This test just verifies that the schema doesn't change, which I think is less valuable than something that explicitly checks the schema. If we want to have a test that validates the schema, can we please simply hard code the expected Arrow schema and compare it to what comes out rather than using this fingerprinting method

We could improve this test in a follow on PR too

tpcdsgen-arrow/src/conversions.rs
```diff
@@ -16,6 +16,26 @@ pub fn decimal_to_i128(d: Decimal) -> i128 {
     d.get_number() as i128
 }
 
+#[deprecated(note = "use a fixed TPC-DS decimal array helper")]
```
alamb: since we aren't offering API guarnatees we can probably just remove the methods rather than deprecating them 🤔

tpcdsgen-arrow/src/tables/web_site.rs
```diff
@@ -198,7 +196,7 @@ fn make_schema() -> SchemaRef {
         Field::new("web_market_manager", DataType::Utf8View, true),
         Field::new("web_company_id", DataType::Int32, true),
         Field::new("web_company_name", DataType::Utf8View, true),
-        Field::new("web_street_number", DataType::Int32, true),
+        Field::new("web_street_number", DataType::Utf8View, true),
```
alamb, line 204: it does appear that the canonical schema is char for this column 🤷 

https://github.com/databricks/tpcds-kit/blob/1b7fb7529edae091684201fab142d956d6afd881/tools/tpcds.sql#L307

alamb (approved):
Makes sense to me -- thank you @mwc360 an @clflushopt

tpcdsgen-arrow/tests/schema.rs
mwc360: updated

tpcdsgen-arrow/src/conversions.rs
mwc360: updated

tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -95,7 +95,7 @@ struct ParquetArgs {
     #[arg(short = 'c', long, default_value = "SNAPPY")]
     compression: Compression,
 
-    /// Target row-group size in bytes
+    /// Approximate target size in uncompressed row-group bytes in Parquet files
```
kevinjqliu, line 98: nit: i recently change this to align the help text of both tpch and tpcds (#424). I'll follow up to align them again

tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -828,6 +827,29 @@ fn test_tpcgen_cli_tpcds_csv_single_table() {
...
+        .assert()
+        .success();
+
+    let contents = fs::read_to_string(temp_dir.path().join("web_returns.csv"))
+        .expect("Failed to read CSV file");
+    let header = contents.lines().next().expect("CSV output is empty");
+    assert!(header.contains(",wr_account_credit,"));
+    assert!(!header.contains("wr_store_credit"));
+}
+
```
kevinjqliu:
```suggestion
```
nit: i feel like this test alone is unnecessary. we can test all the tpcds tables together

tpcdsgen-arrow/tests/schema/expected.rs
```diff
@@ -0,0 +1,489 @@
+//! Explicit expected Arrow schemas for the TPC-DS tables.
+
+use arrow::datatypes::{DataType, Field, Schema, TimeUnit};
+use tpcdsgen::config::Table;
+
+pub(super) fn expected_schema(table: Table) -> Schema {
```
alamb, line 6: 👍

alamb (approved):

alamb:
Thank you very muhch @mwc360

tpcgen-cli/tests/cli_integration/tpcds.rs
kevinjqliu, line 1: nit: i think we can remove both of these tests
- `test_tpcgen_cli_tpcds_csv_uses_canonical_web_returns_header`
- `test_tpcgen_cli_tpcds_parquet_uses_canonical_schemas`

because we now have `tpcdsgen-arrow/tests/schema.rs` to validate the schemas

tpcdsgen-arrow/tests/schema.rs
```diff
@@ -16,6 +12,94 @@ use tpcdsgen_arrow::{
     TimeDimArrow, WarehouseArrow, WebPageArrow, WebReturnsArrow, WebSalesArrow, WebSiteArrow,
 };
 
+#[path = "schema/expected.rs"]
+mod expected;
+use expected::expected_schema;
+
+fn table_schemas(session: &Session) -> Vec<(Table, SchemaRef)> {
```
kevinjqliu, line 19: with #447, we can simplify this and get rid of `Session`

tpcdsgen-arrow/tests/schema.rs
```diff
@@ -593,6 +677,256 @@ fn canonical_schemas(session: &Session) -> Vec<(&'static str, SchemaRef, &'stati
...
+        assert_eq!(
+            csv_header(table, ',').expect("CSV header"),
+            arrow_header,
+            "{table_name}"
+        );
+    }
+}
+
+#[test]
+fn integer_widths_match_lakebench_v4() {
```
kevinjqliu, line 703: nit: this one and some of the other tests are redundant now that we're asserting the declared schema in `expected.rs`

kevinjqliu (commented):
Thanks for the fix @mwc360. I have a few nits from reviewing. I'll follow up with a PR 😄

PR #447, opened 2026-09-16
Use Arrow schema accessors for column validation

kevinjqliu:
Uses `schema_ref()` from #427 in the TPC-H and TPC-DS column-encoding helpers instead of constructing generators and Arrow readers just to inspect schemas.

This removes the need to pass a TPC-DS session or TPC-H scale factor into schema lookups. It also restores TPC-H's distributions and text-pool warmup to its pre-#405 position, after output planning and just before table generation. Column matching and encoding behavior are unchanged.


tpcgen-cli/src/tpcds_cli/parquet.rs
kevinjqliu, line 1: remove `session` now that its not necessary when creating schema

tpcgen-cli/src/tpch_cli/generator.rs
```diff
@@ -236,20 +231,16 @@ impl Default for GeneratorConfig {
     }
 }
 
-/// Returns `table`'s Arrow schema. Does not generate any rows.
-///
-/// `part` and `part_count` do not change the schema, so this always asks
-/// for `(1, 1)`.
-pub(super) fn table_schema(table: Table, scale_factor: f64) -> SchemaRef {
+pub(super) fn table_schema(table: Table) -> SchemaRef {
```
kevinjqliu, line 234: `scale_factor` is no longer needed to get schema

tpcgen-cli/src/tpch_cli/generator.rs
```diff
@@ -336,23 +325,12 @@ impl TpchGenerator {
...
-        // table. validate_column_encodings (below) builds a real generator
-        // per table to read its schema, and every generator also creates
-        // these statics. Warm up first, or the cost hides inside
-        // validation and this timing is wrong.
-        let start = Instant::now();
-        Distributions::static_default();
-        TextPool::get_or_init_default();
-        let elapsed = start.elapsed();
-        info!("Created static distributions and text pools in {elapsed:?}");
-
```
kevinjqliu, line 349: Revert this change that was made in #405, it was necessary when session was still needed to validate column encoding.

We dont need to do this anymore now that `session` is no longer used, so moving it back to its original place

See
https://github.com/datafusion-contrib/tpcgen-rs/pull/405/changes#diff-afaaf72378630b442ef3f73e0c79fe47934ecf7b1c1391d86e8c8aa9b3e4b49eR334-R372

kevinjqliu (commented):
Remove the need to provide `session` to create schema. 

Also move the text pool warmup code to after the column encoding validation code. This allow invalid column name fail a lot faster, from **831 ms to 6.4 ms**.

alamb (approved):

alamb:
Thanks @kevinjqliu  -- this is great

PR #454, opened 2026-09-18
perf(tpcds): speed up weighted distribution lookups

kevinjqliu:
## What changed

Weighted distribution lookups currently use `linear search`. This PR replaces them with `binary search`, which is safe because [`WeightsBuilder::compute_and_add_next_weight`](https://github.com/datafusion-contrib/tpcgen-rs/blob/19c620165cde0783871fc74f04727551fadacafa/tpcdsgen/src/distribution/utils.rs#L37-L43) builds nondecreasing cumulative weights.

`pick_random_index_from_weights` already used `binary_search`; this PR moves it to the same `partition_point` lookup so duplicate cumulative weights return the first matching index.

## Benchmark

Command:

```shell
hyperfine --warmup 2 --runs 5 \
  --prepare "rm -rf /tmp/output" \
  "tpcgen-cli-main tpcds parquet --scale-factor=100 --tables=item --num-threads=1 --compression=UNCOMPRESSED --output-dir=/tmp/output" \
  "tpcgen-cli-branch tpcds parquet --scale-factor=100 --tables=item --num-threads=1 --compression=UNCOMPRESSED --output-dir=/tmp/output"
```

This branch (933 ms):

```text
Time (mean ± σ):     933.2 ms ± 59.6 ms    [User: 904.9 ms, System: 20.8 ms]
Range (min … max):   888.0 ms … 1037.5 ms    5 runs
```

Main (1.03 s):

```text
Time (mean ± σ):      1.025 s ± 0.036 s    [User: 0.982 s, System: 0.025 s]
Range (min … max):    0.989 s … 1.083 s    5 runs
```

For SF100 `item` generation, this branch is **1.10x faster**.


kevinjqliu:
ok i couldn't stop the urge to squeeze out some extra performance gains 😆

alamb (approved):
Makes sense to me -- thank you @kevinjqliu

PR #456, opened 2026-09-19
perf(tpcds): borrow strings in Arrow adapters

kevinjqliu:
## What changed

TPC-DS Arrow adapters currently clone row strings into temporary `Vec<Option<String>>` buffers, then `StringViewBuilder` copies the same strings again into Arrow-owned storage.

This PR keeps those intermediate values borrowed as `Option<&str>`, including static `Y`/`N` values, until Arrow performs the required copy. `date_dim` and `time_dim` likewise borrow their row string fields instead of cloning them. The generated row batches remain alive through array construction, so no unsafe code or lifetime extension is needed.

This removes 84 `.to_owned()` calls and 8 equivalent string clones across 17 adapters without changing schemas, null bitmaps, sentinel handling, or output. Formatting-only strings and address clones remain unchanged.

## Benchmark

Both binaries were built from the same `main` revision with:

```shell
cargo build --locked --release -p tpcgen-cli
```

Command:

```shell
hyperfine --warmup 2 --runs 5 \
  --prepare "rm -rf /tmp/output" \
  "/tmp/tpcgen-cli-main tpcds parquet --scale-factor=100 --tables=item --num-threads=1 --compression=UNCOMPRESSED --output-dir=/tmp/output --quiet" \
  "/tmp/tpcgen-cli-branch tpcds parquet --scale-factor=100 --tables=item --num-threads=1 --compression=UNCOMPRESSED --output-dir=/tmp/output --quiet"
```

This branch (1.05 s):

```text
Time (mean ± σ):      1.048 s ± 0.029 s    [User: 0.949 s, System: 0.029 s]
Range (min … max):    1.013 s … 1.087 s    5 runs
```

Main (1.10 s):

```text
Time (mean ± σ):      1.099 s ± 0.024 s    [User: 0.992 s, System: 0.030 s]
Range (min … max):    1.076 s … 1.136 s    5 runs
```

For single-threaded, uncompressed SF100 `item` generation, this branch is **1.05x faster**, with wall time reduced by **4.6%** and user CPU reduced by **4.3%**. The generated Parquet output is byte-identical to `main`.


tpcdsgen-arrow/src/tables/call_center.rs
```diff
@@ -77,33 +77,33 @@ impl Iterator for CallCenterArrow {
         }
 
         let mut cc_sk: Vec<Option<i32>> = Vec::with_capacity(rows.len());
-        let mut cc_id: Vec<Option<String>> = Vec::with_capacity(rows.len());
+        let mut cc_id: Vec<Option<&str>> = Vec::with_capacity(rows.len());
```
alamb, line 80: 👍

alamb (approved):
This is great -- thank you @kevinjqliu  -- every 5% helps. Nice. This thing will be singing before we know it

Issue #308, opened 2026-06-30
[TCP-DS] Update all references to tpcdsgen and tpchgen-cli to say `tpcgen`


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

We are in the process of making a single binary (`tpcgen`) that makes both tpc-h and tpch-ds datasets intsead of `tpchgen-cli` and `tpcds`.
- https://github.com/clflushopt/tpchgen-rs/issues/293

As @kevinjqliu  pointed out in https://github.com/clflushopt/tpchgen-rs/pull/296#issuecomment-4836196870 some of the readme and help text still may refer to the old binary name(s). We should update it all to refer to the new name

A possible exception is when we invoke the `tpchgen-cli` command (which uses the same code) as `tpcgen` -- it would be nice if the messages there still use tpchgen-cli if possible

clflushopt:
We should also update the README to reflect this

Issue #299, opened 2026-06-29
[TPC-DS] Update main readme to reflect new binaries / TPC-DS feature


alamb:
- part of https://github.com/clflushopt/tpchgen-rs/issues/218

The main README currently describes tpchgen_cl: https://github.com/clflushopt/tpchgen-rs/blob/5a6b7da1076d9a8d7af24edd5f72268396b0acf5/README.md#L7-L6

As we are now adding TPC-DS features we should update the README:
1. Note `tpcgen` makes both TPC-H and TPC-DS datasets
2. Examples using it
3. Note that `tpchgen_cli` still is offered, but only does TPC-H
4. Ideally some performance measurements (when we have that)

kevinjqliu:
we should grep `tpchgen-cli` and make sure to replace the specific string when approach or make it not be specific to the CLI name 

context: https://github.com/clflushopt/tpchgen-rs/pull/296#issuecomment-4836196870

PR #458, opened 2026-09-20
docs: update tpcgen-cli docs and help text to use the new binary names

alamb:
- Part of #308
- The main README is deliberately untouched — that is #299

## Rationale

Now that `tpcgen-cli` generates both TPC-H and TPC-DS data, several docs, help strings, and rustdoc comments still referred to `tpchgen-cli`


alamb:
I need to review this / update

PR #460, opened 2026-09-20
refactor: move `RowIter` to `tpcdsgen`, add `GeneratedRow::table()`

alamb:
- Part of #339

1. Moves the `RowIter` adaptor into  `tpcdsgen::row` so it can be reused
2. Adds GeneratedRow::table() to filter only those rows that are needed


tpcdsgen-arrow/src/lib.rs
```diff
@@ -33,71 +30,3 @@ pub use tables::{
 
 /// Default number of rows per [`RecordBatch`](arrow::array::RecordBatch).
 pub const DEFAULT_BATCH_SIZE: usize = 8_000;
-
-/// Adapts a [`RowGenerator`] into a streaming [`Iterator`] of [`GeneratedRow`]s.
```
alamb, line 37: Moved into tpcdsgen as this is not arrow specific

kevinjqliu (approved):

PR #462, opened 2026-09-20
feat: generate each TPC-DS returns table as its own output

alamb:
~Draft as it builds on https://github.com/datafusion-contrib/tpcgen-rs/issues/339~

- Part of #339

## The problem

The DAT and CSV outputs generate a `returns` (e.g `sales_return`) table as a *side effect* of making the `sales` table: one pass over the sales generator opens two writers and sends each emitted row to whichever file it belongs to.

This means that `--tables store_returns` silently rewrites `store_sales.dat`  and it also doesn't fit into the parallel generaiton model where one chunk writes into one one buffer at a time. 

Thus, this PR changes the code to cenerate each table on its own, by walking its generator and keeping only that table's rows — this is what the Parquet and Arrow generators do too.

## Performance

Note this is about ~20% slower at scale factor 1 (DAT 7.0s → 8.5s, CSV 7.2s → 8.4s), because a sales table and its returns table now walk the generator once each.

Note this is already how the Parquet output works, and I hope to improve parallel generation in future PRs (minimize extra work for creating chained returns)


tpcgen-cli/scripts/tpcds/compare-table.sh
```diff
@@ -92,18 +92,6 @@ log_diff() {
     echo -e "${YELLOW}[DIFF]${NC} $*"
 }
 
-# Returns tables are generated by their parent sales generators.
```
alamb, line 95: workaround that is no longer necessary

tpcgen-cli/src/tpcds_cli/generate.rs
```diff
@@ -112,219 +106,102 @@ pub(super) fn generate_table<O: TableOutput>(
...
+///
+/// Progress is counted in source rows; see [`super::progress::register_table`]
+fn generate_rows<G: RowGeneratorFactory, O: TableOutput>(
     output: &O,
     table: Table,
     session: &Session,
-    progress: TableProgress,
+    progress: ProgressHandle,
 ) -> Result<()> {
-    let TableProgress::Single(progress) = progress else {
```
alamb, line 209: This PR also simplifies a bunch of the complexity around the progress reporting of multiple tables

tpcgen-cli/src/tpcds_cli/progress.rs
```diff
@@ -282,45 +227,4 @@ mod tests {
         parts[2].complete();
         assert_eq!(completions.load(Ordering::Relaxed), 1);
     }
-
-    #[test]
-    fn share_across_parts_gates_sales_and_returns_independently() {
```
alamb, line 287: This is part of the simplification now that we only generate one table at a time

kevinjqliu (approved):
LGTM!

kevinjqliu:
I've bene taking a look at how to best split out `StoreSalesRowGenerator` into 2 generators, one for sales and another for returns

PR #461, opened 2026-09-20
refactor: extract TPC-DS plan/schedule into `tpcds_cli::runner`

alamb:
- Part of #339
- Split off of  #389 (you can see how everything fits together in that PR)

`Parquet::generate_tables` has a bunch of generic TPC-DS work to plan how to create the tables in parallel. This PR refactors that shared code into structures that can be reused for CSV and DAT generation

PR #459, opened 2026-09-20
refactor: move shared generation plumbing to the tpcgen-cli crate root

alamb:
- Part of #339
- Split off of  #389 (you can see how everything fits together in that PR)

This is a refactor to move some of the plumbing currently used in tpchgen-cli to a location where it can be reused by tpcgen as well

PR #463, opened 2026-09-20
perf: parallelize TPC-DS DAT and CSV generation

### LINKED ISSUE #339: [TPC-DS] Multi-core generation of csv/dat

alamb:
Draft as it builds on  #459, #460, #461 and #462.** Once those merge, this will be rebased on main and its diff will collapse to just the parallel generation work.

- Part of #339
- closes https://github.com/datafusion-contrib/tpcgen-rs/pull/389
- closes https://github.com/datafusion-contrib/tpcgen-rs/issues/339

## The change

The DAT and CSV outputs used a single core. Moar cores is better.

They now work the way the TPC-H text outputs and TPC-DS Parquet already do: split each table into chunks of source rows, format the chunks into in-memory buffers on separate threads, and have one writer task append the buffers to the file in order. Tables are generated concurrently within the overall thread budget, largest first.

DAT and CSV differ only in how a chunk of rows is formatted into a buffer, which is what the new `RowFormat` trait captures. That leaves all three TPC-DS outputs on one path — `plan_tables` / `run_plans` to schedule, `generate_in_chunks` to generate — and that path is shared with TPC-H.

## Performance

Main

main

This branch
branch

Following the @kevinjqliu methodology in https://github.com/datafusion-contrib/tpcgen-rs/pull/456

Both binaries were built from the same `main` revision with:

```shell
cargo build --locked --release -p tpcgen-cli
```

Command:

```shell
hyperfine --warmup 2 --runs 5   --prepare "rm -rf /tmp/output"  \
"/tmp/tpc-parallel/target/release/tpcgen-cli tpcds --scale-factor=10  --output-dir=/tmp/output --quiet"   \
"/tmp/tpc-main/target/release/tpcgen-cli tpcds --scale-factor=10 --output-dir=/tmp/output --quiet"
```

## Scale Factor 10 (8x 🕐  faster 🚀 )
The overall generation time is 8x faster
```shell
Benchmark 1: /tmp/tpc-parallel/target/release/tpcgen-cli tpcds --scale-factor=10  --output-dir=/tmp/output --quiet
  Time (mean ± σ):      7.869 s ±  0.358 s    [User: 101.378 s, System: 2.409 s]
  Range (min … max):    7.436 s …  8.220 s    5 runs

Benchmark 2: /tmp/tpc-main/target/release/tpcgen-cli tpcds --scale-factor=10 --output-dir=/tmp/output --quiet
  Time (mean ± σ):     63.021 s ±  0.120 s    [User: 60.934 s, System: 1.642 s]
  Range (min … max):   62.837 s … 63.129 s    5 runs
```

Note however, that it uses more CPU overall time (101s vs 61s), because it re-has to generate the large fact tables (e.g. `store_sales`) when generating the return tables (e.g. `store_returns`) and then discards the fact table rows immediately. Fixing this is tracked by
- https://github.com/datafusion-contrib/tpcgen-rs/issues/398


tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -60,6 +65,15 @@ enum Commands {
 struct DatArgs {
     #[command(flatten)]
     common: CommonArgs,
+
+    /// The number of threads for parallel generation, defaults to the number of CPUs
```
kevinjqliu: nit: this is now in all 3 (dat/csv/parquet), maybe we can move it to `CommonArgs`

tpcgen-cli/src/tpcds_cli/plan.rs
```diff
@@ -147,7 +180,7 @@ impl IntoIterator for TpcdsGenerationPlan {
 /// generate the group. Sales and returns must both use their paired sales
 /// table's source-row count, not their output-row count. Fractional bytes avoid
 /// large rounding errors for narrow tables such as inventory.
-fn estimated_bytes_per_source_row(table: Table) -> f64 {
+fn estimated_parquet_bytes_per_source_row(table: Table) -> f64 {
```
kevinjqliu, line 175: these estimates are based off of sf100 (documented in the comment above) 

calling it out because csv/dat estimate is using sf1

tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -241,86 +269,62 @@ impl CommonArgs {
...
-                let mut table_sessions = Vec::with_capacity(tables.len() * parts.len());
-                for table in &tables {
-                    let sessions = parts
-                        .iter()
-                        .map(|&part| self.to_session(Some(table.get_name().to_string()), part))
-                        .collect::<Result<Vec<_>>>()?;
-                    // One bar per table, shared across all its parts.
-                    let table_progress = output.register_table(*table, &sessions, progress.clone());
-                    let part_progress = share_handle_across_parts(table_progress, sessions.len());
-                    for (session, progress) in sessions.into_iter().zip(part_progress) {
```
kevinjqliu, line 316: amazing!

tpcgen-cli/src/tpcds_cli/progress.rs
```diff
@@ -1,43 +1,11 @@
...
-
-/// Register one progress bar for `table`.
-///
-/// `sessions` holds one session per requested part. Callers register once, then
-/// split the result with [`share_handle_across_parts`].
-///
-/// Progress is counted in *source* rows (e.g. `store_sales` rather than
-/// `store_returns`) rather than output rows, as only the source rows are known
-/// exactly up front.
-pub(super) fn register_table(
```
kevinjqliu, line 30: i like the new pattern of using `share_handle_across_parts` and calling `complete` per worker. The progress tracker can figure out how much progress is made.

tpcgen-cli/src/tpcds_cli/plan.rs
```diff
@@ -1,42 +1,57 @@
-//! [`TpcdsGenerationPlan`]: row group layout for TPC-DS Parquet files.
+//! [`TpcdsGenerationPlan`]: how a TPC-DS table is split into chunks.
 
 use std::ops::RangeInclusive;
 use tpcdsgen::config::Table;
 
 /// Parquet files can have at most 32767 row groups
 const MAX_ROW_GROUPS: u64 = 32767;
```
kevinjqliu: nit: maybe move to parquet

kevinjqliu (approved):
LGTM! 

My agent flagged an issue with generation that's not introduced by this PR. More details in https://github.com/datafusion-contrib/tpcgen-rs/issues/457#issuecomment-5764906851

EDIT: looks like the problem is actually introduced by this PR 😄 (https://github.com/datafusion-contrib/tpcgen-rs/issues/457#issuecomment-5765213392)

alamb:
I filed a ticket to track the incorrect results here
- https://github.com/datafusion-contrib/tpcgen-rs/issues/475

Seems to be like parquet already has the problem. This PR basically just makes the same problem happen for DAT/CSV (which isn't great I realize)

I suggest we merge this PR and then fix all three formats as a follow on

kevinjqliu:
> I suggest we merge this PR and then fix all three formats as a follow on

SGTM!

tpcgen-cli/src/tpcds_cli/plan.rs
kevinjqliu: Filed an issue for this as a follow up:

https://github.com/datafusion-contrib/tpcgen-rs/issues/476

kevinjqliu:
Nice! Lets get this moving! I have a fix for the TCPDS SCD tables in #478

PR #464, opened 2026-09-20
Test ranged TPC-DS sales and returns generation

kevinjqliu:
## Summary

- extend the Parquet ranged-vs-single-pass regression to catalog, store, and web sales and returns
- use small row groups and verify every output crosses a range boundary
- compare each ranged result with the corresponding single-pass Arrow reader

Stacked on #462, which makes sales and returns independent outputs.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p tpcgen-cli`


kevinjqliu:
mostly superseded by #462

Issue #457, opened 2026-09-19
[FEATURE] TPCDS Add more correctness coverage for multi-part generation for all tables at SF10


alamb:
**Is your feature request related to a problem? Please describe.**
While working on this PR https://github.com/datafusion-contrib/tpcgen-rs/pull/406

It was very hard to test that generation in parts was equivalent to a single part because it takes so long to generate the data and tpcds only splits tables into parts when they have more than 1m rows. 

I instead settled for some spot checks at scale factor 1

It would be better to do a full verificiation of the data

Once we have implemented multi core generation, I think the speed will be etter
- https://github.com/datafusion-contrib/tpcgen-rs/issues/339

**Describe the solution you'd like**

I suggest creating a conformance test that runs tpcgen tpcds with mulitple parts at scale factor maybe 5 or 10, concatenates the results into a single file, and then compares that with the md5 sum of the entire dataset. 

This will avoid having to re-generate SF10 with the old generators, and should be fairly fast once we get parallel generation going 

**Describe alternatives you've considered**

**Additional context**

alamb:
Note on which tables actually get split, as it determines the scale factor this test needs.

Only tables with at least 1,000,000 **source** rows are split across parts (`SMALL_TABLE_ROW_THRESHOLD` in `tpcdsgen/src/config/session.rs`); everything else is written entirely to part 1, so at scale factor 1 a `--parts` run only really exercises two tables:

| scale factor | tables that split (source rows) |
| --- | --- |
| 1 | `customer_demographics` (1,920,800), `inventory` (11,745,000) |
| 5 | + `store_sales` / `store_returns` (1,200,000) |
| 10 | + `catalog_sales` / `catalog_returns` (1,600,000) |

`web_sales` / `web_returns` (600,000 source rows at SF10) need SF ≥ 17 and stay unsplit either way.

Two reasons to prefer **SF10 over SF5**:

* SF5 misses `catalog_sales` / `catalog_returns` entirely.
* `MD5SUMS` are committed for both `scale-10-trino` and `scale-10-c`, while SF5 only has `scale-5-c` (no Trino hashes).

And per the "Known coverage limits" section of `TESTING.md`, no workflow currently checks the scale-10 `MD5SUMS` at all — so this test would need no fixture regeneration, exactly as hoped for above, and would start exercising hashes that are already in the repo.

### Measured locally (prototype, not committed)

[...]

PR #474, opened 2026-09-21
test: reproduce TPC-DS SCD range regression

kevinjqliu:
> [!NOTE]
> **Superseded by #478. This was never merged into `main`.**
>
> This PR was the standalone reproducer. Its base was temporarily set to the fix branch (`kevinjqliu-fix-tpcds-scd-range-history`) so the test could be validated against the fix. Pushing that branch made these commits reachable from their own base, so GitHub auto-marked this PR as merged into it.
>
> The "Merged" badge therefore refers to the fix branch, not `main`. No code reached `main` through this PR.
>
> The reproducer test below now lives in **#478** alongside the fix, where it passes. Review and merge #478 instead.

Relates to #475

## What

Adds `item` to the existing test that compares Parquet output with a single uninterrupted generator run. The existing `store_sales` and `store_returns` coverage stays the same.

## Why

Each Parquet row group is generated independently from its own source row range. The generator skips to the first row in that range by advancing the RNG streams as if the earlier rows had been generated.

That works for regular tables because each row only depends on its row number and RNG state. It is not enough for SCD tables.

An Item can have multiple physical rows for the same logical item. The first row starts the item, and later versions copy some values from the previous version while changing others. Skipping advances the RNG state, but it does not generate the earlier rows, so `previous_row` is never populated.

## Why Item fails here

```text
SF 0.001 item rows:             2,000
Estimated size:                 2,000 * 188.84 = 377,680 bytes
Row groups:                     ceil(377,680 / 250,000) = 2
Rows per group:                 ceil(2,000 / 2) = 1,000
Second group starts at row:     1,001
```

Item rows follow this repeating six-row layout:

```text
row % 6 = 1          one-version Item
row % 6 = 2, 3       two versions of the same Item
row % 6 = 4, 5, 0    three versions of the same Item
```

Row 1,000 has remainder 4, so it creates the first version of an Item. Row 1,001 has remainder 5, so it is the second version of that same Item and needs values from row 1,000.

The second row group starts at row 1,001 with the RNG in the right position, but row 1,000 was skipped instead of generated. There is no previous Item row to copy from, so the result differs from uninterrupted generation. The test fails today and should pass once the previous SCD row is rebuilt when starting a range.

## Original dsgen doesn't have this issue

Original `dsgen` doesn't split generation below 1M rows. `item` tops out at 502K rows at SF100K, so it is never partitioned and this case is never reached.

Tpcgen splits smaller datasets into Parquet row groups or output chunks, which is why we see the issue here. See [the discussion in #457](https://github.com/datafusion-contrib/tpcgen-rs/issues/457#issuecomment-5765213392) for more context.


alamb (approved):

alamb:
Perhaps we should file a bug explaining the problem? The existing ticket covers just adding more tests (at least that is the implication from the summary)
- https://github.com/datafusion-contrib/tpcgen-rs/issues/457

alamb:
I will file the ticket

alamb:
Filed 
- https://github.com/datafusion-contrib/tpcgen-rs/issues/475

kevinjqliu:
Yea it would be helpful to file a new issue and capture the nuance of the probme (esp since it differs from the original dsgen implementation). 
I can file a ticket too if you're busy

kevinjqliu:
I added more details to https://github.com/datafusion-contrib/tpcgen-rs/issues/475 😄

kevinjqliu:
Sadly Claude messed this PR state up 😭. This is not "Merged" as displayed by Github.

> [!NOTE]
> **Superseded by #478. This was never merged into `main`.**
>
> This PR was the standalone reproducer. Its base was temporarily set to the fix branch (`kevinjqliu-fix-tpcds-scd-range-history`) so the test could be validated against the fix. Pushing that branch made these commits reachable from their own base, so GitHub auto-marked this PR as merged into it.
>
> The "Merged" badge therefore refers to the fix branch, not `main`. No code reached `main` through this PR.
>
> The reproducer test below now lives in **#478** alongside the fix, where it passes. Review and merge #478 instead.

PR #360, opened 2026-07-08
[tpcgen-cli]: parallel TPC-DS parquet generation (10x faster)

alamb:
## Which issue does this PR close?

- Closes #334
- part of #218

## Rationale

I want `tpcgen-cli tpcds parquet` to be fast and to do so, we need to use all cores, the same as we do for TPCH

This PR a planning one logical chunk per row group and encoding row groups concurrently with the shared writer (#342). This PR applies the same model to TPC-DS.

## What changes

- Add `TpcdsGenerationPlan` to compute the layout of the file up front
- Complete the source row range API from #346
- Replace the TPC-DS `ArrowWriter` path with the shared parquet writer that was extracted in https://github.com/clflushopt/tpchgen-rs/pull/342

Note the shared writer now embeds the Arrow schema in the parquet metadata (as `ArrowWriter` does), so Arrow types with no exact parquet equivalent round trip (e.g. `dbgen_version`'s `Time32(Second)` column). This also makes the resulting parquet files slightly larger (by a constant amount -- the embedded arrow schema)

# Performance at SF 10

Performance is more than 10x faster  on my mac:

```shell
 hyperfine --runs 5 --prepare "rm -rf /tmp/output" "target/release/tpcgen-cli tpcds parquet --scale-factor=10 --tables=store_sales --output-dir /tmp/output"
```

This branch (2.87 s)

```
Benchmark 1: target/release/tpcgen-cli tpcds parquet --scale-factor=10 --tables=store_sales --output-dir /tmp/output
  Time (mean ± σ):      2.876 s ±  0.036 s    [User: 37.710 s, System: 0.738 s]
  Range (min … max):    2.825 s …  2.925 s    5 runs
```

Main (30s)

```
Benchmark 1: target/release/tpcgen-cli tpcds parquet --scale-factor=10 --tables=store_sales --output-dir /tmp/output
  Time (mean ± σ):     30.397 s ±  0.062 s    [User: 30.048 s, System: 0.326 s]
  Range (min … max):   30.319 s … 30.466 s    5 runs
```

PR #478, opened 2026-09-21
fix: rebuild TPC-DS SCD history when starting a source row range

### LINKED ISSUE #475: [TPC-DS] Incorrect results with mutli-part generation

As @kevinjqliu  shows in this reproducer
- https://github.com/datafusion-contrib/tpcgen-rs/pull/474

At some scales TPC-DS generates incorrect results for scaled tables due to how the data is split into pieces.

We need to fix this before release

## Details

Root cause:

Each output chunk or Parquet row group is generated independently from its own source row range. To start a range, we advance the RNG streams as if the earlier rows had been generated.

That works for regular tables, but **not for slowly changing dimension tables**. An Item can have multiple physical rows for the same logical item. Later versions copy some values from the previous version, which is stored in `previous_row`. Skipping advances the RNG state but does not generate the earlier rows, so `previous_row` is never rebuilt.

This is a partition-boundary issue, not a threading issue. It reproduces with `--num-threads 1` because the ranges are still generated independently.

### Example

#474 reproduces this with Parquet at SF 0.001:

```text
Item rows:                       2,000
Estimated size:                 2,000 * 188.84 = 377,680 bytes
Row groups:                     ceil(377,680 / 250,000) = 2
Rows per group:                 ceil(2,000 / 2) = 1,000
Second group starts at row:     1,001
```

Item rows follow this repeating layout:

```text
row % 6 = 1          one-version Item
row % 6 = 2, 3       two versions of the same Item
row % 6 = 4, 5, 0    three versions of the same Item
```

Row 1,000 creates the first version of an Item. Row 1,001 is the second version of that same Item and needs values from row 1,000. The second range starts with the RNG in the right position, but row 1,000 was skipped instead of generated, so the output differs from a single uninterrupted pass.

## Scope

The same problem affects all SCD tables:

- `item`
- `store`
- `web_page`
- `web_site`
- `call_center`

`item`, `store`, and `web_page` can silently produce different output. `call_center` can return an uninitialized previous-row error, and `web_site` can panic.

Original `dsgen` does not hit this through normal partitioning because it does not split tables below 1M rows. `item` tops out at 502K rows at SF100K, so this case is never reached. See [[the dsgen comparison in #457](https://github.com/datafusion-contrib/tpcgen-rs/issues/457#issuecomment-5765213392)](https://github.com/datafusion-contrib/tpcgen-rs/issues/457#issuecomment-5765213392).

## Expected behavior

Generating and concatenating source-row ranges should produce the same output as one uninterrupted generator run. Starting an SCD range needs to rebuild the previous version state before generating the first requested row.

kevinjqliu:
- Closes #475, which has the full root cause writeup
- Includes the reproducer test from #474

## The change

Starting an SCD source row range now rebuilds the previous version state before generating the first requested row, so `item`, `store`, `web_page`, `web_site` and `call_center` no longer depend on rows the range skipped.

- Skipping to a source row clears `previous_row`, so stale state can never be reused.
- Before generating the first row of a range, replay the versions this row inherits from. The six-row revision cycle bounds that at two rows.
- The replay lives in one helper in `slowly_changing_dimension_utils` instead of being repeated in each of the five generators.

The Arrow reparse test used to pick a skip starting row that always landed on a new business key, stepping around this bug. That workaround is removed, so its skip tests now start mid-version.

#360 introduced this when it gave each Parquet row group its own source row range. #463 puts DAT and CSV on the same chunked path, so fixing it in the generators covers all three outputs.

## Tests

`test_tpcgen_cli_tpcds_parquet_matches_single_pass_generation` now also generates `item` with `--row-group-bytes` small enough to force several source row ranges, and compares it against a single uninterrupted run. It fails on `main` with 12 rows wrong and passes here.

The generators also get unit coverage walking every position of the six-row cycle, plus a seek test covering forward, backward, and repeated seeks.

## Benchmark

No measurable cost at default row group sizes. The worst case is 5%, generating only SCD tables with 100 KB row groups so the replay runs as often as possible (1.257 s on `main` vs 1.316 s here, 567 source row ranges over 300,000 `item` rows). A full TPC-DS SF10 run is unchanged.


tpcdsgen-arrow/tests/reparse.rs
```diff
@@ -208,28 +208,6 @@ where
...
-    }
-
-    // SCD tables reuse values from the previous source row for continuation
-    // records. Pick a new business-key row so a skipped generator does not need
-    // previous-row state initialized before the first generated row.
-    (1..=max_skip_row)
-        .rev()
-        .find(|row| row % 6 == 1)
-        .unwrap_or(1)
-}
```
kevinjqliu, line 231: interestingly enough, we were previously avoiding this issue by finding a row with that does not require previous revisions ([`row % 6 == 1`](https://github.com/datafusion-contrib/tpcgen-rs/blob/9ed8401b425b0210bff7128c878224bf75823db8/tpcdsgen/src/slowly_changing_dimension_utils.rs#L57-L62))

tpcdsgen/src/row/tables/call_center_row_generator.rs
```diff
@@ -418,6 +424,9 @@ impl RowGenerator for CallCenterRowGenerator {
     fn skip_rows_until_starting_row_number(&mut self, starting_row_number: u64) {
         self.abstract_generator
             .skip_rows_until_starting_row_number(starting_row_number);
+        // Invalidate the retained SCD state.
+        // This tells generate_row_and_child_rows to replay it when needed.
+        self.previous_row = None;
```
kevinjqliu: same change for all the `tables/**`

this is the best inline comment i can come up with to explain the situation. suggestions are welcomed

tpcdsgen/src/slowly_changing_dimension_utils.rs
```diff
@@ -108,6 +111,48 @@ pub fn compute_scd_key(table: Table, row_number: u64) -> SlowlyChangingDimension
     SlowlyChangingDimensionKey::new(business_key, start_date, end_date, is_new_key)
 }
 
+/// How many rows before `row_number` have to be replayed to rebuild its history.
+///
+/// The six-row cycle below is the one [`compute_scd_key`] uses to assign business
+/// keys, so the count reaches back to the row that begins this entity.
+fn previous_rows_needed(row_number: u64) -> u64 {
```
kevinjqliu, line 149: this matches [compute_scd_key](https://github.com/datafusion-contrib/tpcgen-rs/blob/9ed8401b425b0210bff7128c878224bf75823db8/tpcdsgen/src/slowly_changing_dimension_utils.rs#L51-L102)

tpcgen-cli/tests/cli_integration/tpcds.rs
```diff
@@ -1041,7 +1044,7 @@ fn test_tpcgen_cli_tpcds_parquet_matches_single_pass_generation() {
         .arg("--scale-factor")
         .arg("0.001")
         .arg("--tables")
-        .arg("store_sales,store_returns")
+        .arg("store_sales,store_returns,item")
```
kevinjqliu, line 1042: includes the reproducer test from #474

kevinjqliu:
Since #463 is now merged, I'll create a follow up PR (stacked perhaps) that will test SCD tables for DAT/CSV. Similar to `test_tpcgen_cli_tpcds_parquet_matches_single_pass_generation`

tpcdsgen/src/row/tables/call_center_row_generator.rs
alamb: Maybe you could also leave a link to the github issue (https://github.com/datafusion-contrib/tpcgen-rs/issues/475) for more information

> // Invalidate the retained slowly changing dimension (SCD) state to ensure 
> // correct start row. See https://github.com/datafusion-contrib/tpcgen-rs/issues/475

tpcdsgen/src/row/row_iter.rs
```diff
@@ -163,4 +166,57 @@ mod tests {
         let rows = rows_for(ReasonRowGenerator::new, Table::Reason, &session, &[(1, 0)]);
         assert!(rows.is_empty());
     }
+
+    /// Assert that generating `table` one source row at a time reproduces the
+    /// unranged output. Every row is a range start, so this covers each
```
alamb, line 171: the 6 row revision cycle is kind of confusing here I think without the context

Maybe we could explain briefly that slow changing dimension (scd) tables change their value every 6 values of the base table (is that correct?)

perhaps we could put that explanation in a comment of `tpcdsgen/src/slowly_changing_dimension_utils.rs` (or link to the existing explanation) and then point this comment there

tpcdsgen/src/slowly_changing_dimension_utils.rs
```diff
@@ -108,6 +111,48 @@ pub fn compute_scd_key(table: Table, row_number: u64) -> SlowlyChangingDimension
...
+pub(crate) fn generate_scd_history<G: RowGenerator>(
+    generator: &mut G,
+    row_number: u64,
+    session: &Session,
+) -> Result<()> {
+    let previous_rows = previous_rows_needed(row_number);
+    if previous_rows == 0 {
+        return Ok(());
+    }
+    let first_revision = row_number - previous_rows;
```
alamb, line 181: Suggested invariant check

```rust
let first_revision = row_number - previous_rows;
debug_assert_eq!(
    previous_rows_needed(first_revision), 0,
    "replay must start on a new business key or generate_row_and_child_rows recurses"
);
```

alamb (approved):
Thank you @kevinjqliu  -- this makes sense and the test speaks for itself

The only major suggestion I have is to document this behavior / learning somewhere in the comments so our future selves (or others reading the code) will have a better change of understanding the nuance

Perhaps a comment in the tpcdsgen/src/slowly_changing_dimension_utils.rs module would be the best place. We could do this as a follow on PR too


Here are locations that it is partly described found by my agent
<details>
1. What an SCD table is — two sentences, buried in a private join-key helper:
   https://github.com/datafusion-contrib/tpcgen-rs/blob/6508ba06280ebd23713a34be81a8c62b1df2eb80/tpcdsgen/src/join_key_utils.rs#L257-L260
2. The six-row revision cycle — only as terse // 2 of 2 revisions labels on the match arms of compute_scd_key; it never states that the cycle is 6 rows or that rows 3/5/6 copy from earlier rows:
   https://github.com/datafusion-contrib/tpcgen-rs/blob/6508ba06280ebd23713a34be81a8c62b1df2eb80/tpcdsgen/src/slowly_changing_dimension_utils.rs#L51-L106
3. How fields are carried forward (field_change_flag bit mechanics, plus the "bug in the C code" note) — explained once, in call_center only:
   https://github.com/datafusion-contrib/tpcgen-rs/blob/6508ba06280ebd23713a34be81a8c62b1df2eb80/tpcdsgen/src/row/tables/call_center_row_generator.rs#L119-L130
</details>

kevinjqliu:
Thanks for the suggestions @alamb. I improved the documentation as much as possible. Let me know what you think 😄

tpcdsgen/src/row/tables/call_center_row_generator.rs
```diff
@@ -418,6 +425,10 @@ impl RowGenerator for CallCenterRowGenerator {
     fn skip_rows_until_starting_row_number(&mut self, starting_row_number: u64) {
         self.abstract_generator
             .skip_rows_until_starting_row_number(starting_row_number);
+        // Invalidate the retained slowly changing dimension (SCD) state.
```
alamb, line 428: 👍

tpcdsgen/src/slowly_changing_dimension_utils.rs
```diff
@@ -1,4 +1,37 @@
...
+//!
+//! The first revision of an entity generates its values. Every later revision
+//! reuses the same business key, and
+//! [`get_value_for_slowly_changing_dimension`] copies forward the fields that
+//! do not change, so each generator keeps the row it generated last.
+//!
+//! Generating a range of source rows fast forwards the random number streams
+//! to the first row of the range. The skipped rows are never generated. A
+//! range that starts on a later revision has no previous revision to copy
+//! from, so `generate_scd_history` replays the earlier ones to restore the
```
alamb, line 26:
```suggestion
//! from, so [`generate_scd_history`] replays the earlier ones to restore the
```

alamb (approved):
love it -- thank you

tpcdsgen/src/slowly_changing_dimension_utils.rs
kevinjqliu: unfortunately this will break build because `generate_scd_history` is `pub(crate)`, and module docs are public output.

```
error: public documentation for `slowly_changing_dimension_utils` links to private item `generate_scd_history`
   = note: `-D rustdoc::private-intra-doc-links` implied by `-D warnings`
error: could not document `tpcdsgen`
```

alamb:
Nice!

PR #479, opened 2026-09-22
test: cover TPC-DS CSV source row ranges for SCD tables

kevinjqliu:
Adds the CSV counterpart to `test_tpcgen_cli_tpcds_parquet_matches_single_pass_generation`. #463 gave the text outputs their own chunked implementation, with no test that chunked output matches a single uninterrupted pass.

Generates `item` through the CLI and compares against `ItemArrow`. `item` is a slowly changing dimension table, and scale factor 3.2 splits it into two source row ranges at row 19,001, a continuation revision.

Needs #478 to pass; validated locally against it. Red on `main` until that merges.


kevinjqliu:
we should wait for #480 to land. Having `--chunk-bytes` will make the test so much simpler

kevinjqliu:
Closing in favor of #483

PR #480, opened 2026-09-22
Configure TPC-DS text chunk size

kevinjqliu:
Adds `--chunk-bytes` to TPC-DS DAT and CSV generation. The default remains 8 MiB, and changing it does not change the generated output.

Also reuses the existing positive byte validation used by Parquet row-group sizing.

Tested with `cargo test -p tpcgen-cli` and Clippy.


tpcgen-cli/README.md
```diff
@@ -19,3 +19,22 @@ tpcgen-cli tpch parquet -s 1 --tables lineitem --column-encoding=l_comment=DELTA
 tpcgen-cli tpcds csv -s 1 --output-dir /tmp/tpcds
 tpcgen-cli tpcds csv -s 1 --delimiter='\t' --output-dir /tmp/tpcds
 ```
+
+## TPC-DS text chunk sizing
```
alamb, line 23: I am not sure that this needs a documentation page entry unless a lot f people end up tuning the chunksize -- I don't expect it to be very common

alamb (commented):
Thanks @kevinjqliu  -- I am not sure we should make this parameter configurable as I don't think it does what we think

I looked into this and I think this constant is actually only used to calculate the row counts per chunk when planning eg https://github.com/datafusion-contrib/tpcgen-rs/blob/1e06be003af4b255c0427b725237a37901879314/tpcgen-cli/src/tpcds_cli.rs#L232


The actual generation chunk size is hard coded to 8mb deep in the bowels:
https://github.com/datafusion-contrib/tpcgen-rs/blob/1cb58ac1974f410d45beb0c34161b0459f043087/tpcgen-cli/src/generate.rs#L88-L87

Thus, I would actually suggest we move the constant into the module it is used in and don't pass it in to Dat and Csv as a parameter

kevinjqliu:
Yea I was debating whether this config is useful. The original motivation is for testing in #479.

`--chunk-bytes` can help to partition source rows so we can force SCD tables to be chopped up in a way to reproduce the revision bug (see https://github.com/datafusion-contrib/tpcgen-rs/pull/479/changes#diff-06a71db9082f2e711a53095ab76dc80c1d7d53e0d61767a9753492161a6ab83aR1078-R1091) 

This mirrors how we tested the parquet side with `--row-group-bytes` (see https://github.com/datafusion-contrib/tpcgen-rs/blob/8ef70445b02c8da0ee28424435a64565d7ef6466/tpcgen-cli/tests/cli_integration/tpcds.rs#L1043-L1045) 

Conceptually I think `--chunk-bytes` is useful to control the peak memory usage. ie setting it to 1MB instead of the default 8MB produces more smaller chunks. With that said, I dont think there's any benefits to set a bigger size 

Heres a local run for fun:
| Chunk target | Planned chunks | Median wall | Throughput | Median user CPU | Median peak RSS |
| --- | --- | --- | --- | --- | --- |
| 1 MiB | ~191 | 0.29s | 630.8 MiB/s | 1.12s | 17.8 MiB |
| 8 MiB | ~24 | 0.29s | 630.8 MiB/s | 1.10s | 71.1 MiB |
| 16 MiB | ~12 | 0.30s | 609.7 MiB/s | 1.11s | 131.8 MiB |
| 32 MiB | ~6 | 0.38s | 481.4 MiB/s | 1.08s | 192.5 MiB |
| 64 MiB | ~3 | 0.40s | 457.3 MiB/s | 1.09s | 192.3 MiB |
| 128 MiB | ~2 | 0.56s | 326.6 MiB/s | 1.06s | 192.1 MiB |

kevinjqliu:
The run is interesting because it looks like 1MB is overall better than 8MB. Same wall time and throughput but using less peak RSS 👀 

EDIT: tracking this in https://github.com/datafusion-contrib/tpcgen-rs/issues/485

kevinjqliu:
I think #483 is a better test for the CSV/DAT than #479, which means we can close both #479 and this PR 😄 

less is more!

PR #481, opened 2026-09-22
minor: Update TPC-DS chunk size estimates for scale factor 100

alamb:
- Follow on to #463
- prompted by @kevinjqliu 's comment in https://github.com/datafusion-contrib/tpcgen-rs/pull/463#discussion_r4063561025, which pointed out that the DAT/CSV estimates were measured at a different scale factor than the Parquet ones.

## What

`plan.rs` sizes each generation chunk from an estimated output bytes per *source* row. Those estimates were measured three different ways:

* Parquet: scale factor 100, but the large tables were sampled from single row groups rather than measured over the whole file.
* DAT and CSV: scale factor 10.

I had my agent measures all three with whole files at scale factor 100 so they are consistent

## DAT and CSV

The text estimates were low on tables that get split into many chunks:

| table (DAT) | before (SF10) | after (SF100) | |
| --- | ---: | ---: | ---: |
| `inventory` | 20.76 | 21.60 | +4.1% |
| `web_returns` | 170.09 | 175.59 | +3.2% |
| `store_returns` | 140.98 | 145.16 | +3.0% |
| `store_sales` | 1666.22 | 1706.65 | +2.4% |
| `web_sales` | 2519.04 | 2577.25 | +2.3% |
| `catalog_sales` | 1898.68 | 1938.53 | +2.1% |

The small dimension tables were already exact — `date_dim`, `time_dim`, `customer_demographics`, `ship_mode` and `income_band` . CSV tracks DAT within a few tenths of a percent throughout.

## Parquet

These numbers are very similar. I regenerated the full scale factor 100 dataset with 128 MiB row groups and recomputed `sum(total_uncompressed_size) / source_rows` for all 25 tables. The largest deviation from the committed values was `customer` at +1.25% (76.65 → 77.61); `catalog_sales` was +0.22%, `store_sales` +0.16%, and most tables matched exactly.

## `dbgen_version`

Keeps its previous value in all three tables. Its "size" is just the command line it records


tpcgen-cli/src/tpcds_cli/plan.rs
```diff
@@ -176,31 +176,31 @@ fn estimated_parquet_bytes_per_source_row(table: Table) -> f64 {
     match table {
         Table::CallCenter => 229.80,
         Table::CatalogPage => 108.21,
-        Table::CatalogReturns => 65.97,
-        Table::CatalogSales => 668.72,
-        Table::Customer => 76.65,
+        Table::CatalogReturns => 66.08,
```
alamb, line 179: All the numbers are very close

kevinjqliu (approved):
LGTM!

PR #482, opened 2026-09-22
feat: support `tpcgen tpcds --stdout`

### LINKED ISSUE #336: [TPC-DS] Support generating output to stdout

`tpcgen tpch` currently supports writing output to stdout. `tpcgen tpcds` should too

```
      --stdout
          Write the output to stdout instead of a file
```

<details><summary>Output of <pre>tpcgen tpch</pre> </summary>
<p>

```shell
TPC-H data

Usage: tpcgen-cli tpch [OPTIONS]
       tpcgen-cli tpch <COMMAND>

Commands:
  tbl      Generate TBL (pipe-delimited) output
  csv      Generate CSV output with CSV-specific options
  parquet  Generate Apache Parquet output with Parquet-specific options
  help     Print this message or the help of the given subcommand(s)

Options:
  -s, --scale-factor <SCALE_FACTOR>
          Scale factor to create

          [default: 1]

  -o, --output-dir <OUTPUT_DIR>
          Output directory for generated files (default: current directory)

          [default: .]

  -T, --tables <TABLES>
          Which tables to generate (default: all)

          Possible values:
          - region:   Region table (alias: r)
          - nation:   Nation table (alias: n)
          - supplier: Supplier table (alias: s)
          - customer: Customer table (alias: c)
          - part:     Part table (alias: P)
          - partsupp: PartSupp table (alias: S)
          - orders:   Orders table (alias: O)
          - lineitem: LineItem table (alias: L)

  -p, --parts <PARTS>
          Number of part(itions) to generate. If not specified creates a single file per table

      --part <PART>
          Which part(ition) to generate (1-based). If not specified, generates all parts

  -n, --num-threads <NUM_THREADS>
          The number of threads for parallel generation, defaults to the number of CPUs

          [default: 16]

  -v, --verbose
          Verbose output

          When specified, sets the log level to `info` and ignores the `RUST_LOG` environment variable. When not specified, uses `RUST_LOG`

  -q, --quiet
          Quiet mode - only show error-level logs

      --stdout
          Write the output to stdout instead of a file

      --no-progress
          Disable progress bars during data generation.

          Bars are also auto-suppressed by `--quiet`, `--stdout`, or when stderr is not a terminal.

  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version
```

</p>
</details>

alamb:
# Which issue does this PR close?

- Closes #336

# Rationale for this change

`tpcgen tpch` can write generated data to stdout with `--stdout`, which makes it easy to pipe data straight into another tool rather than staging it on disk. `tpcgen tpcds` had no such option. 

# What changes are included in this PR?

This makes the TPC-DS CLI behave the same as the TPC-H CLI.

Example:

```shell
$ tpcgen-cli tpcds --stdout -s 0.001 -T reason
1|AAAAAAAABAAAAAAA|R1|Package was damaged                               |
...
```

# Are these changes tested?
Yes

# Are there any user-facing changes?


tpcgen-cli/src/tpch_cli/output_plan.rs
```diff
@@ -11,30 +13,6 @@ use std::fmt::{Display, Formatter};
 use std::io;
 use std::path::PathBuf;
 
-/// Where a partition will be output
```
alamb, line 14: refactored into common output_location

tpcgen-cli/src/tpcds_cli/csv.rs
```diff
@@ -29,15 +29,19 @@ use tpcdsgen::row::GeneratedRow;
 /// CSV output generator.
 #[derive(Debug, Clone)]
 pub(super) struct Csv {
-    output_dir: PathBuf,
+    base_location: OutputLocation,
```
alamb, line 32: this PR centralizes the PathBuf/Stdout stuff into `OutputLocation` for both TPC-DS and TPC-H

tpcgen-cli/src/tpcds_cli/generate.rs
```diff
@@ -183,10 +186,20 @@ where
         generator: PhantomData,
     });
 
-    generate_file(&path, sources, num_threads, progress.clone()).await?;
+    match &location {
```
alamb, line 189: here is one place where the new codepath is used (there are equivalents for csv and dat)

tpcgen-cli/src/tpcds_cli.rs
```diff
@@ -309,14 +317,30 @@ impl CommonArgs {
...
     ) -> (
         Arc<dyn ProgressTracker>,
         Option<Box<dyn io::Write + Send + 'static>>,
     ) {
+        // Show progress only on an interactive terminal and when no flag
+        // suppresses it. `--stdout` is included so piped data isn't
+        // interleaved with bar redraws on shared shells.
         #[cfg(feature = "indicatif-progress")]
-        if self.progress_bars_enabled && !self.quiet && io::stderr().is_terminal() {
+        if self.progress_bars_enabled && !self.quiet && !self.stdout && io::stderr().is_terminal() {
```
kevinjqliu, line 343: 👍 aligns with tpch 

https://github.com/datafusion-contrib/tpcgen-rs/blob/db342f5db6120a97ee509011d5b676b44b374d7b/tpcgen-cli/src/tpch_cli/cli.rs#L196

tpcgen-cli/src/output_location.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use std::fmt::{Display, Formatter};
+use std::io;
+use std::path::{Path, PathBuf};
+
+/// Where generated data is written: the filesystem, or stdout.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum OutputLocation {
+    /// Output to a file in the specified directory
+    File(PathBuf),
```
kevinjqliu, line 11: nit: `File` can represent both directory and file, which can be difficult to keep track. ie calling `create_dir_all` will create a folder even if the path name is `reason.dat`

just calling it out, i dont think we need to over engineer this

kevinjqliu (approved):
LGTM!

One callout: `--stdout` may interleave output from multiple generators in the same stream. For example, `--tables item,reason` writes both tables to stdout, and concurrent generation can mix their chunks. This matches the existing `tpch` CLI behavior, so I don’t think this PR needs to address it. 🤷

alamb:
> One callout: --stdout may interleave output from multiple generators in the same stream. For example, --tables item,reason writes both tables to stdout, and concurrent generation can mix their chunks. This matches the existing tpch CLI behavior, so I don’t think this PR needs to address it. 🤷

Yes this is a good call.  My primary usecase for this is to do CPU profiling (so it isn't bottlenecked on IO)

We should probably add some documentation that explains this behavior in the CLI

alamb:
> We should probably add some documentation that explains this behavior in the CLI

- Filed https://github.com/datafusion-contrib/tpcgen-rs/issues/484

kevinjqliu:
merge queue failed on a github api error (hard to be github these days 😭)

> Error: getaddrinfo ENOTFOUND api.github.com


Lets try again

PR #483, opened 2026-09-22
test: verify multi-part TPC-DS generation for dat, csv and parquet

### LINKED ISSUE #457: [FEATURE] TPCDS Add more correctness coverage for multi-part generation for all tables at SF10

alamb:
Closes #457.

## Rationale

Multi-part generation had no end-to-end coverage: `dsdgen` only ever partitions
tables with at least 1,000,000 rows and the checked-in `MD5SUMS` describe whole
tables, so nothing verified that concatenating `--parts N` output reproduces a
single-pass run. #457 proposed doing exactly that — generate with several parts,
concatenate, compare to the MD5 of the whole dataset.

This also closes a gap `TESTING.md` called out: `MD5SUMS` are committed for
`scale-10-trino` and `scale-10-c`, but no workflow checked them.

## What changes

**`compare-table.sh` grows two flags**, following the existing pattern:

- `--parts N` — generate the table as N parts (`<table>/<table>.<i>.<ext>`),
  concatenate them in part order, and compare the result as if it had been
  generated in one pass.
- `--format dat|csv|parquet` — which output format to check.

```console
$ ./scripts/tpcds/compare-table.sh customer_demographics --parts 4
[INFO] =========================================
[INFO] Table Comparison: customer_demographics (dat, MD5-only, 4 parts)
[INFO] =========================================
[INFO] Trino MD5SUMS: .../tests/fixtures/tpcds/scale-1-trino/MD5SUMS
[INFO] Generating customer_demographics with Rust...
[INFO] Generated 4 nonempty part(s) of 4 requested
[INFO] Rust output: 1920800 rows
[INFO] Trino expected MD5: 8831872c6d56ea9d4f24701f2feaef48
[INFO] Rust MD5:              8831872c6d56ea9d4f24701f2feaef48
[SUCCESS] ✓ customer_demographics: MD5 match (1920800 rows, 8831872c6d56ea9d4f24701f2feaef48)
[INFO] =========================================
```

What each format is compared against:

| format | reference |
| --- | --- |
| `dat` | the committed `MD5SUMS` (or the `.dat` fixture with `--full`) |
| `csv` | a single-file CSV run of the same binary |
| `parquet` | a single-file Parquet run of the same binary, row by row |

`csv` and `parquet` have no reference implementation of their own, so all they
can be compared against is a single-file run — which is only meaningful when the
output under test was split, hence both require `--parts`. The `.dat` suite is
what anchors them to the reference implementations, per the chain of trust in
`TESTING.md`.

`compare-all-tables.sh` passes both flags through.

**`tpcgen-cli/examples/parquet-compare.rs`** is new. Parquet files cannot be
concatenated with `cat`, so this reads the ordered parts as one row stream,
compares it against the single-file output value by value, and reports the first
row that differs:

```console
MISMATCH: row 0 (0-based) differs, first in column 'cd_demo_sk'
expected:
+------------+-----------+--- ...
| 1          | M         | ...
actual:
+------------+-----------+--- ...
| 480201     | M         | ...
```

**`.github/workflows/tpcds-parts-conformance.yml`** is new: all 24 tables, 10
parts, scale factor 10, one job per format (`dat` in both compat modes, `csv`
and `parquet` in `trino`). It runs on `merge_group` and `workflow_dispatch`
rather than per PR — `workflow_dispatch` takes `scale` and `parts` inputs so it
can be pointed at other settings by hand.

Peak disk was a concern at SF10 (`store_sales.dat` is 3.7 GB), so the MD5-only
path streams the parts into `md5` instead of materializing the concatenation;
only `--full`, which needs a file to `diff`, writes one, and it deletes each part
as it appends.

## Caveats

- Because tables under 1,000,000 source rows are never split, only six tables
  genuinely exercise this at SF10: `store_sales`, `store_returns`,
  `catalog_sales`, `catalog_returns`, `inventory` and `customer_demographics`.
  The other 18 — `web_sales` and `web_returns` among them — come out as a
  single part. `TESTING.md` now says so under "Known coverage limits".
- **This passes on `main` today, where I expected it to fail until #478.** All
  four matrix legs are green locally on this branch:

  | leg | result | wall clock |
  | --- | --- | --- |
  | `dat`, trino | 24/24 | 42s |
  | `dat`, c | 24/24 | 41s |
  | `csv`, trino | 24/24 | 377s |
  | `parquet`, trino | 24/24 | 35s |

  The SCD tables #475 is about (`item`, `store`, `call_center`, `web_page`,
  `web_site`) also pass at SF10 with and without `--parts 10` — at that scale
  they are all far under the 1,000,000-row split threshold, so `--parts` never
  splits them and this job does not reach that bug. Covering it would need a
  scale factor where those tables cross the threshold, or a lower threshold.
  Happy to extend the matrix if that is what you want out of this.

  (`csv` is the slow leg because it generates twice — parts plus the
  single-file reference — over the largest uncompressed output of the three.
  The job's timeout is 90 minutes, so there is plenty of headroom.)

## Testing

Locally, against this branch:

- The full CI matrix at its real settings (SF10, 10 parts, all 24 tables, all
  four legs) — see the table above.
- `compare-table.sh` with and without `--parts`, for all three formats, at SF1.
- Negative controls on `parquet-compare`: a missing trailing part is reported as
  a row-count difference, and parts fed out of order are reported as a value
  difference at row 0.
- Argument validation: `--format csv` without `--parts`, `--parts 0`, an unknown
  `--format`, and `--full` with a non-`dat` format all fail with a specific
  message.
- `cargo fmt --check` and `cargo clippy --all-targets --all-features` are clean
  (the new example is covered by both, since clippy uses `--all-targets`).

Draft because the new workflow has not run in CI yet (it only triggers on
`merge_group` and `workflow_dispatch`, so a PR run will not exercise it).


alamb:
Notes for myself on my test machine that is generating different scale factors from the c code

Working from https://github.com/alamb/tpcds-data

<details>
ssh alamb@34.28.250.40

alamb@alamb-tpcds:~/tpcds-data/tools$ mkdir sf1
alamb@alamb-tpcds:~/tpcds-data/tools$ ./dsdgen -DIR sf1 -SCALE 1
dsdgen Population Generator (Version 4.0.0)
Copyright Transaction Processing Performance Council (TPC) 2001 - 2021

I have a bunch of those going for all the different scale factors 1-10

i will revalidate the md5 sums from known good ones on https://github.com/alamb/tpcds-data and then also fill in the other sizes


Then run like

alamb@alamb-tpcds:~/tpcds-data/tools$ bash -c 'for scale in 1 2 3 ; do echo "# Scale Factor $scale"; echo "\`\`\`"; md5sum sf$scale/* ; echo "\`\`\`"; echo ; done ' | tee  ~/shasum.md

</details>

alamb:
Superceded by 
- https://github.com/datafusion-contrib/tpcgen-rs/pull/487
- https://github.com/datafusion-contrib/tpcgen-rs/pull/488

